PR #2, opened 2020-04-20
trusted-firmware-m: scripts: add required `bytes` cast when signing

microbuilder:
This commit adds an explicit cast to `bytes` for the payload
when signing data with imgtool. On some systems, the data may be
provided as a `ByteArray`, which will cause the signing function
to fail since the underlying library expects `bytes`.

Signed-off-by: Kevin Townsend <kevin.townsend@linaro.org>


ioannisg:
@microbuilder are you also sending an upstream patch for this change?

microbuilder:
@ioannisg I can send one upstream as well, I just wanted to get this fix in here while I'm testing the recent update to TF-M 1.0 for Zephyr.  You can't build without this fix on some platforms.

PR #3, opened 2020-05-03
cmake: Fixes to CMake to ensure Zephyr can build TF-M using ninja

tejlmand:
This PR fixes dependency issues when reviewing https://github.com/zephyrproject-rtos/zephyr/pull/19985

---------
This commit fixes the following CMake issues in TF-M build system:
- armgcc depfile corrected (was using clang flags)
- mbedcrypto byproduct added
- s_veneers.o byproduct added

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/trusted-firmware-m/cmake/Common/CompilerGNUARMCommon.cmake
+++ b/trusted-firmware-m/cmake/Common/CompilerGNUARMCommon.cmake
@@ -146,6 +146,7 @@ function(compiler_set_cmse_output TARGET FILE_PATH)
 	set_property(TARGET ${TARGET} APPEND_STRING PROPERTY LINK_FLAGS " -Wl,--cmse-implib,--out-implib=${FILE_PATH}")
 	#Tell cmake cmse output is a generated object file.
 	SET_SOURCE_FILES_PROPERTIES("${FILE_PATH}" PROPERTIES EXTERNAL_OBJECT true GENERATED true)
+	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_COMMAND} -E echo "" BYPRODUCTS ${FILE_PATH})
 	#Tell cmake cmse output shall be removed by clean target.
 	get_directory_property(_ADDITIONAL_MAKE_CLEAN_FILES DIRECTORY "./" ADDITIONAL_MAKE_CLEAN_FILES)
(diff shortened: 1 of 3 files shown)


tejlmand:
@galak @karl-zh This fixes a couple of deps issues I discovered while looking into https://github.com/zephyrproject-rtos/zephyr/pull/19985

tejlmand:
note: it's just a draft as there could be more things to patch.
Secondly, this is not fully tested.
Just the basic discoveries to get TF-M to build with ninja in Zephyr context.

trusted-firmware-m/cmake/Common/CompilerGNUARMCommon.cmake
```diff
@@ -146,6 +146,7 @@ function(compiler_set_cmse_output TARGET FILE_PATH)
 	set_property(TARGET ${TARGET} APPEND_STRING PROPERTY LINK_FLAGS " -Wl,--cmse-implib,--out-implib=${FILE_PATH}")
 	#Tell cmake cmse output is a generated object file.
 	SET_SOURCE_FILES_PROPERTIES("${FILE_PATH}" PROPERTIES EXTERNAL_OBJECT true GENERATED true)
+	add_custom_command(TARGET ${TARGET} POST_BUILD COMMAND ${CMAKE_COMMAND} -E echo "" BYPRODUCTS ${FILE_PATH})
```
galak, line 149: is this just debug?

tejlmand: No, it's because the `${FILE_PATH}` which is the `s_veneers.o` file is created as part of the final linking.
But the linking, which happens because of the `add_executable()` does not support the `BYPRODUCTS` keyword.

This is done manually by controlling the properties on the files, and using `add_dependencies`, which works well for `make` but `ninja` is a bit more strict, and want to know a specific target that can create the file.

Therefore `add_custom_command()` is used, just to make `ninja` happy.
It doesn't need to do anything, but it provides the needed relation between the file and the target.

tejlmand:
> Can you add a `trusted-firmware-m: ` prefix to the commit message. Trying to use that so we can easily tell any local changes.

Done.

PR #5, opened 2020-05-06
trusted-firmware-m: scripts: add python requirements.txt

microbuilder:
This commit adds a requirements.txt file with the python
requirements for TF-M when signing images, etc.

Signed-off-by: Kevin Townsend <kevin.townsend@linaro.org>

### DIFF
--- a/trusted-firmware-m/bl2/ext/mcuboot/scripts/requirements.txt
+++ b/trusted-firmware-m/bl2/ext/mcuboot/scripts/requirements.txt
@@ -1,0 +1,4 @@
+cryptography
+pyasn1
+pyyaml
+cbor>=1.0.0


galak:
Is this requirements some how different than what mcuboot would require?  I assume the signing here is for mcuboot's purpose, so feel like these requirements should come from mcuboot and not TFM.

microbuilder:
> Is this requirements some how different than what mcuboot would require? I assume the signing here is for mcuboot's purpose, so feel like these requirements should come from mcuboot and not TFM.

They are also used by some of the utilities in the `tools` folder, for ex.:

```
./tools/tfm_parse_manifest_list.py:12:from jinja2 import Environment, BaseLoader, select_autoescape
./tools/tfm_parse_manifest_list.py:45:        http://jinja.pocoo.org/docs/2.10/api/#jinja2.BaseLoader.get_source
```

d3zd3z:
Given that the goal is to get rid of the mcuboot within TFM, why don't we do:

1. put the requirements.txt file from upstream mcuboot into the mcuboot within tf-m.
2. Have a TFM-specific requirements.txt file that adds the additional requirements needed for tfm, and than has an include directive to pull in the ones from MCUboot. Look at `scripts/requirements.txt` in a Zephyr tree to see how to do this. I think the path is relative to the directory of the current file.

microbuilder:
@d3zd3z We reference the mcuboot requirements.txt file in CI. Does that mean we should instead reference the new file in the TF-M repo from CI, rather than Juul's one? It's unlikely MCUBoot will change or that we'll have a requirement, but it does seem odd to use the BL2 requirements when Zephyr relies on MCUBoot itself.

d3zd3z:
I think it would make sense to have a top-level TFM one, referenced from CI.

microbuilder:
@d3zd3z For now, I've moved this into `trusted-firmware-m/bl2/ext/mcuboot/scripts` since the img signing script is required for TF-M builds to pass in CI, but I can do an analysis after the 2.3 merge window closes for a generic `requirements.txt` of all the dependencies of this repo, and include the bl2 file as part of that.

carlescufi:
@microbuilder doesn't upstream tf-m come with its own `requirements.txt`?

microbuilder:
> @microbuilder doesn't upstream tf-m come with its own `requirements.txt`?

Doesn't seem like it, no, they're just documented.

PR #11, opened 2020-08-18
stm32l5: flash: modify script to flash zephyr

NolwennV:
In the current implementation for STM32L5 MCUs, the binaries path and filenames in the script TFM_update.sh are hardcoded.
This is a minimal modification to be able to flash in zephyr, as the scripts are likely to change after the integration of the new build system.

### DIFF
--- a/trusted-firmware-m/platform/ext/target/stm/stm32l5xx/boards/scripts/TFM_UPDATE.sh
+++ b/trusted-firmware-m/platform/ext/target/stm/stm32l5xx/boards/scripts/TFM_UPDATE.sh
@@ -34,12 +34,12 @@ nvcounter=
 boot=0x0c001000
 unused=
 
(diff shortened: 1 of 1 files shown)


microbuilder (approved):

microbuilder:
These changes will also need to happen upstream in the TF-M repository or they will be lost on a future update. I would also suggest reconsidering some of the filenames for these scripts to be more descriptive, such as `TFM_UPDATE.sh` --> `flash.sh`, but this should take place upstream. As an interim solution, this at least avoids the issue of having to manually change the files, though.

microbuilder:
@NolwennV I've merged these changes in, but as part of https://github.com/zephyrproject-rtos/zephyr/pull/27524 you'll need to update the `west.yaml` file's TF-M entry to point to the updated SHA for the TF-M repo: `7b763c0dc3bd3600f7e73aecca51d82a45cf2e35`.

PR #12, opened 2020-08-25
CMakeLists.txt: make BL2 configurable

agansari:
Allow boards to set BL2 variable when building TFM. The scope of this
change is to allow developers to build images without MCUboot.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ cmake_minimum_required(VERSION 3.13.1)
 # CFGFILE: The TF-M config file to use, without the .cmake extension
 # IPC: Build TFM IPC library. This library allows a non-secure application to
 #      interface to secure domain using IPC.
+# BL2: Boolean if build uses MCUboot.
 #
 # The following output values can also be used:
 #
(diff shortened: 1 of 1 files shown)


agansari:
@microbuilder @d3zd3z @galak

microbuilder:
This should default to `True` so as not to break any existing calls to this function, which previously assumed usage of BL2.

agansari:
> This should default to `True` so as not to break any existing calls to this function, which previously assumed usage of BL2.

@microbuilder i've set the variable as True by default.

microbuilder:
@tejlmand Is this the right way to handle default values in a function, in your opinion?

CMakeLists.txt
```diff
@@ -61,7 +67,7 @@ function(trusted_firmware_build)
     BINARY_DIR ${TFM_BINARY_DIR}
     CMAKE_ARGS -DPROJ_CONFIG=${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/configs/${TFM_CFGFILE}.cmake
                -DTARGET_PLATFORM=${TFM_BOARD}
-               -DBL2=True
+               -DBL2=${TFM_BL2}
```
tejlmand, line 70: Only some config default to: `BL2=True`, like:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/master/trusted-firmware-m/configs/ConfigDefaultProfileS.cmake#L37
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/master/trusted-firmware-m/configs/ConfigPsaApiTestIPC.cmake#L51

Others are not defaulting to true, so I think it would be better to let `TFM` config decide the default value, and only pass on the value if defined by the user.

```suggestion
               ${TFM_BL2_ARG}
```

```

File: CMakeLists.txt [hunk 2]
```diff
@@ -28,12 +29,17 @@ cmake_minimum_required(VERSION 3.13.1)
...
+#                        BL2 True)
 function(trusted_firmware_build)
   set(options IPC)
-  set(oneValueArgs BINARY_DIR BOARD CFGFILE OUT_VENEERS_FILE OUT_INCLUDE_PATH)
+  set(oneValueArgs BINARY_DIR BOARD CFGFILE OUT_VENEERS_FILE OUT_INCLUDE_PATH BL2)
   cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "" ${ARGN})
 
+  if(NOT DEFINED TFM_BL2)
+    set(TFM_BL2 True) #by default True
+  endif()
```
tejlmand, line 41: See other comment.
```suggestion
  if(DEFINED TFM_BL2)
    set(TFM_BL2_ARG "-DBL2=${TFM_BL2}")
  endif()
```

tejlmand (changes requested):
Let's avoid deciding what default means, as that is already handled in the TFM config files.

CMakeLists.txt
microbuilder: I'm OK with this, but it will require updating the `mps2_an521_nonsecure` and `nucleo_l552ze_q` to set the BL2 status to true:

- https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/tfm_integration/tfm_ipc/CMakeLists.txt#L11
- https://github.com/zephyrproject-rtos/zephyr/blob/master/samples/tfm_integration/psa_level_1/CMakeLists.txt#L10

agansari: @microbuilder there is already BL2 set to True by default in Zephyr's TFM samples.

agansari:
@microbuilder @tejlmand made the changes requested.

microbuilder (approved):

tejlmand (approved):

PR #14, opened 2020-09-30
CMakeLists.txt: Provide common routine for post-build commands

anangl:
... so that those commands are not duplicated in particular boards that
support building with TF-M.

Signed-off-by: Andrzej Głąbek <andrzej.glabek@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ function(trusted_firmware_build)
   endif()
 
   include(ExternalProject)
-  
+
   ExternalProject_Add(
     tfm
     SOURCE_DIR ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m
@@ -92,6 +92,77 @@ function(trusted_firmware_build)
     target_include_directories(tfm_ipc_psa_api PUBLIC ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include)
     target_compile_definitions(tfm_ipc_psa_api PUBLIC TFM_PSA_API TFM_PARTITION_TEST_CORE_IPC)
     target_link_libraries(tfm_ipc_psa_api PRIVATE zephyr_interface)
+  endif()
+
+endfunction()
+
+# Gets a list of commands to be performed after a successful Zephyr build that
+# involves trusted-firmware-m.
+function(trusted_firmware_get_post_build_commands out_commands)
+
+  if(NOT CONFIG_TFM_BL2_FALSE)
+
+    # Set default image versions if not defined elsewhere
+    if(NOT DEFINED TFM_IMAGE_VERSION_S)
+      set(TFM_IMAGE_VERSION_S 0.0.0+0)
+    endif()
+    if(NOT DEFINED TFM_IMAGE_VERSION_NS)
+      set(TFM_IMAGE_VERSION_NS 0.0.0+0)
+    endif()
+
+    set(TFM_MCUBOOT_DIR "${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/bl2/ext/mcuboot")
+    set(PREPROCESSED_FILE "${CMAKE_BINARY_DIR}/tfm/image_macros_preprocessed")
+
+    if(NOT DEFINED TFM_PUBLIC_KEY_FORMAT)
+      set(TFM_PUBLIC_KEY_FORMAT "full")
+    endif()
+
+    set(${out_commands}
+
+      # Sign secure binary image with public key
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/imgtool.py
+        sign
+        --layout ${PREPROCESSED_FILE}_s.c
+        -k ${CONFIG_TFM_KEY_FILE_S}
+        --public-key-format ${TFM_PUBLIC_KEY_FORMAT}
+        --align 1
+        -v ${TFM_IMAGE_VERSION_S}
+        ${ADD_NS_IMAGE_MIN_VER}
+        ${ADD_SECURITY_COUNTER_S}
+        -H 0x400
+        ${CMAKE_BINARY_DIR}/tfm/install/outputs/${TFM_TARGET_PLATFORM}/tfm_s.bin
+        ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+
+      # Sign non-secure binary image with public key
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/imgtool.py
+        sign
+        --layout ${PREPROCESSED_FILE}_ns.c
+        -k ${CONFIG_TFM_KEY_FILE_NS}
+        --public-key-format ${TFM_PUBLIC_KEY_FORMAT}
+        --align 1
+        -v ${TFM_IMAGE_VERSION_NS}
+        ${ADD_S_IMAGE_MIN_VER}
+        ${ADD_SECURITY_COUNTER_NS}
+        -H 0x400
+        --included-header
+        ${CMAKE_BINARY_DIR}/zephyr/${KERNEL_BIN_NAME}
+        ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+
+      # Create concatenated binary image from the two independently signed binary files
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/assemble.py
+        --layout ${PREPROCESSED_FILE}_s.c
+        -s ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+        -n ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+        -o ${CMAKE_BINARY_DIR}/tfm_sign.bin
+
+      # Copy mcuboot.bin
+      COMMAND ${CMAKE_COMMAND} -E copy
+        ${CMAKE_BINARY_DIR}/tfm/install/outputs/${TFM_TARGET_PLATFORM}/mcuboot.bin
+        ${CMAKE_BINARY_DIR}
+
+      PARENT_SCOPE
+    )
+
   endif()
 
 endfunction()


ioannisg (approved):

ioannisg:
@erwango @oyvindronningstad FYI.

This is the tf-m module counterpart patch of #28806. It only groups together the post-build actions when we build a "non-secure" platform with tfm support.

Tested on mps2 and stm32 platforms.

erwango (approved):

microbuilder:
@agansari Do you mind having a look at this for the LPC while I'm out? Thanks!

agansari:
> @agansari Do you mind having a look at this for the LPC while I'm out? Thanks!

Tested on LPC, works OK.

agansari (approved):

oyvindronningstad (approved):

galak (commented):
This feels a little backwards to me, while I get the benefit of having this in common, is it possible that common cmake code can live in Zephyr?

galak:
@tejlmand any thoughts on how this might live purely in the zephyr tree in a common place?  Would it make sense to have modules/tfm.cmake and place this in there?

CMakeLists.txt
```diff
@@ -95,3 +95,74 @@ function(trusted_firmware_build)
...
+      # Sign secure binary image with public key
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/imgtool.py
+        sign
+        --layout ${PREPROCESSED_FILE}_s.c
+        -k ${CONFIG_TFM_KEY_FILE_S}
+        --public-key-format ${TFM_PUBLIC_KEY_FORMAT}
+        --align 1
+        -v ${TFM_IMAGE_VERSION_S}
+        ${ADD_NS_IMAGE_MIN_VER}
+        ${ADD_SECURITY_COUNTER_S}
```
tejlmand, line 131: Those two variables may not be set, that depends on current variables in the scope of the caller, and may lead to unexpected behavior depending on the location from where this function is called.

Probably this should be arguments to the function.

CMakeLists.txt
```diff
@@ -95,3 +95,74 @@ function(trusted_firmware_build)
...
+      # Create concatenated binary image from the two independently signed binary files
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/assemble.py
+        --layout ${PREPROCESSED_FILE}_s.c
+        -s ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+        -n ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+        -o ${CMAKE_BINARY_DIR}/tfm_sign.bin
+
+      # Copy mcuboot.bin
+      COMMAND ${CMAKE_COMMAND} -E copy
+        ${CMAKE_BINARY_DIR}/tfm/install/outputs/${TFM_TARGET_PLATFORM}/mcuboot.bin
```
tejlmand, line 160: The `TFM_TARGET_PLATFORM` variable may not be set, that depends on current scope of the caller.
This may lead to unexpected behavior depending on the location from where this function is called.
Same comment applies to L133.

Probably this should be an argument to the function.

tejlmand (changes requested):

CMakeLists.txt
```diff
@@ -95,3 +95,74 @@ function(trusted_firmware_build)
   endif()
 
 endfunction()
+
+# Gets a list of commands to be performed after a successful Zephyr build that
+# involves trusted-firmware-m.
+function(trusted_firmware_get_post_build_commands out_commands)
```
tejlmand, line 101: Not sure exactly where / which target will call this function, but I think it would be better to have this as a macro which actually adds this as a real post build step, like:
```
macro(sign_trusted_firmware target <other args>)
  if(NOT CONFIG_TFM_BL2_FALSE)
    <existing code>

    add_custom_command(TARGET ${target}
                         POST_BUILD
                          COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/imgtool.py
                          sign
                          <existing additional args>
    )

   <and so on>
  endif()
endmacro()
```

Then users won't need to handle a `out_commands` variable and decide what to do with it.

The reason to suggest a macro here, instead of a function, has to do with the `add_custom_commnad(... POST_BUILD ..)` must be called in same CMakeLists directory as the target.

Also, feel free to find a better name.
Not sure if: `sign_trusted_firmware` is the best name.

If this is purely a Zephyr macro, then maybe it should be called `zephyr_<something>`.

tejlmand (changes requested):
Also, I don't think this function should live in zephyr/CMakeLists.txt file.

Would it not make more sense define this function / macro here:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/master/CMakeLists.txt

tejlmand:
> @tejlmand any thoughts on how this might live purely in the zephyr tree in a common place? Would it make sense to have modules/tfm.cmake and place this in there?

I would like to know why this cannot live in the tfm repo.

tejlmand:
> 
> 
> This feels a little backwards to me, while I get the benefit of having this in common, is it possible that common cmake code can live in Zephyr?

This code only makes sense together with Zephyr.
So I think this is at the right place.

Of course, one could consider to have a `<tfm_module>/cmake/extensions.cmake` for module functions, but then that should be a general principle for all modules.
But I would do such cleanups if we move Zephyr module code into Zephyr, as discussed here:
https://github.com/zephyrproject-rtos/zephyr/pull/27185

CMakeLists.txt
tejlmand: > Not sure exactly where / which target will call this function, but I think it would be better to have this as a macro which actually adds this as a real post build step, like:

I should of course look here:
https://github.com/zephyrproject-rtos/zephyr/pull/28806/files

and that actually means this will be hard to change, due to the nature of:
`extra_post_build_commands`

If changing that, then we are looking into a completely different task.

CMakeLists.txt
anangl: This variable is supposed to be set in the respective board definition, for example in [mps2_an521/board.cmake](https://github.com/zephyrproject-rtos/zephyr/blob/ec9080a03610acd7105d7721432a5ed22c75d98c/boards/arm/mps2_an521/board.cmake#L4):
```
set(TFM_TARGET_PLATFORM "AN521")
```

How about adding here a check like:
```
    if(NOT DEFINED TFM_TARGET_PLATFORM)
      message(FATAL_ERROR "TFM_TARGET_PLATFORM is not set")
    endif()
```

CMakeLists.txt
anangl: This was copied from the mps2_an521 board's [CMakeLists.txt](https://github.com/zephyrproject-rtos/zephyr/blob/12e1fd653deab52cef5b1679ba631c35b2a55286/boards/arm/mps2_an521/CMakeLists.txt#L45-L46). Looking at it now, I don't know how it was supposed to work. Perhaps I should add here:
```
    if (TFM_NS_IMAGE_MIN_VER)
      set(ADD_NS_IMAGE_MIN_VER "-d \"(1,${TFM_NS_IMAGE_MIN_VER})\"")
    else()
      set(ADD_NS_IMAGE_MIN_VER "")
    endif()
    if (TFM_SECURITY_COUNTER_S)
      set(ADD_SECURITY_COUNTER_S "-s ${TFM_SECURITY_COUNTER_S}")
    else()
      set(ADD_SECURITY_COUNTER_S "")
    endif()
```
basing on what is done in MCUBoot.cmake: https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/143df675557305b61f7930a50459a53a8d2bb097/trusted-firmware-m/bl2/ext/mcuboot/MCUBoot.cmake#L127-L131
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/143df675557305b61f7930a50459a53a8d2bb097/trusted-firmware-m/bl2/ext/mcuboot/MCUBoot.cmake#L92-L96

CMakeLists.txt
tejlmand: We are having CMake cached variables and `CONFIG_<name>` Kconfig variables available as globals.
If we start having other variables semi globally available throughout modules, then things becomes much harder to keep track of.

So unless this is Kconfig setting, I would like it to be passed by the caller.

CMakeLists.txt
```diff
@@ -95,3 +95,74 @@ function(trusted_firmware_build)
...
+        --included-header
+        ${CMAKE_BINARY_DIR}/zephyr/${KERNEL_BIN_NAME}
+        ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+
+      # Create concatenated binary image from the two independently signed binary files
+      COMMAND ${PYTHON_EXECUTABLE} ${TFM_MCUBOOT_DIR}/scripts/assemble.py
+        --layout ${PREPROCESSED_FILE}_s.c
+        -s ${CMAKE_BINARY_DIR}/tfm_s_signed.bin
+        -n ${CMAKE_BINARY_DIR}/zephyr_ns_signed.bin
+        -o ${CMAKE_BINARY_DIR}/tfm_sign.bin
```
tejlmand, line 156: I discovered the `tfm_sign.bin` is used here:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/7316751a318fafdd1ae0d3594162dc0a1469157a/CMakeLists.txt#L156

we should not have such a hard coded dependency between module code and CMake code in Zephyr, as that will break if out files are renamed in one place, but not the other.

Instead `tfm_sign.bin` should be returned as an out value (or provided as argument).
Note: I have not checked if there are more such names.

tejlmand (changes requested):

CMakeLists.txt
tejlmand: @microbuilder you took over from @karl-zh, do you have any comments on this ?

tejlmand: For reference: https://github.com/zephyrproject-rtos/zephyr/blame/12e1fd653deab52cef5b1679ba631c35b2a55286/boards/arm/mps2_an521/CMakeLists.txt#L45-L46

galak:
> This code only makes sense together with TFM.
> So I think this is at the right place.
> 
> Of course, one could consider to have a `<tfm_module>/cmake/extensions.cmake` for module functions, but then that should be a general principle for all modules.
> But I would do such cleanups if we move Zephyr module code into Zephyr, as discussed here:
> [zephyrproject-rtos/zephyr#27185](https://github.com/zephyrproject-rtos/zephyr/pull/27185)

There was a general feeling in the process group about trying to have glue code, etc for modules live in the zephyr code base instead of with the module.  I get that the cmake integration isn't exactly glue code, and I've got no issue with it being in the module repo.  But wanted to ask the question.

tejlmand:
> There was a general feeling in the process group about trying to have glue code, etc for modules live in the zephyr code base instead of with the module. I get that the cmake integration isn't exactly glue code, and I've got no issue with it being in the module repo. But wanted to ask the question.

I know, just think that cleanup is outside the scope of this PR, and we should take a look at the exact structure we want in Zephyr for such code.
Some modules could end up with a lot a CMake code in Zephyr, if the end goal is to have modules with no to minimal Zephyr specific code in the module repo.

ioannisg:
@tejlmand @anangl @oyvindronningstad @galak is this old PR still applicable ?

anangl:
> @tejlmand @anangl @oyvindronningstad @galak is this old PR still applicable ?

No. At least the PR here does not make much sense, as the modified code has been moved to the main Zephyr repository.

ioannisg:
> > @tejlmand @anangl @oyvindronningstad @galak is this old PR still applicable ?
> 
> No. At least the PR here does not make much sense, as the modified code has been moved to the main Zephyr repository.

But how about doing this work in the main repository? Is this applicable?

oyvindronningstad:
> 
> 
> > > @tejlmand @anangl @oyvindronningstad @galak is this old PR still applicable ?
> > 
> > 
> > No. At least the PR here does not make much sense, as the modified code has been moved to the main Zephyr repository.
> 
> But how about doing this work in the main repository? Is this applicable?

It's done already

PR #15, opened 2020-10-07
Tfm upmerge to new build system

ioannisg:
Up-merge Upstream TF-M to get the new build system.
Patches:
- a simple **reset --hard** of `./trusted-firmware-m` to the upstream trusted-firmware-m tip of master
- a simple **reset --hard** of `./tf-m-tests` to the upstream tf-m-test tip of master
- the required updates in CMakeLists.txt
- a patch that modifies the toolchain selection

Still need to apply the patch for STM (mentioned in the comments)


ioannisg:
CC @agansari @anangl and @oyvindronningstad

microbuilder:
These path changes will need to be taken into account compared to the upstream TF-M code: https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/7b763c0dc3bd3600f7e73aecca51d82a45cf2e35 ... ideally, we should have a single patch mechanism for this kind of thing if it can't be solved upstream.

ioannisg:
Added a 3rd patch that aims at overwriting the toolchain prefix and path, directly in the GNUARM toolchain cmake file. This would _hopefully_ make CI pass

ioannisg:
> These path changes will need to be taken into account compared to the upstream TF-M code: [7b763c0](https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/7b763c0dc3bd3600f7e73aecca51d82a45cf2e35) ... ideally, we should have a single patch mechanism for this kind of thing if it can't be solved upstream.

is this the _only_ patch that we need to re-apply, @microbuilder ?
any other downstream-only patches that I might have overwritten?

microbuilder:
@ioannisg I believe that's the only one.

d3zd3z:
@ioannisg Have we looked into the git-subtree utility? It's purpose is to do exactly what we're doing here. It ends up with the same results we have, just adding a few lines of metadata to the commits. But, it tracks downstream patches and makes this kind of thing a lot easier to manage.

tejlmand (changes requested):

CMakeLists.txt
```diff
@@ -26,33 +29,58 @@ cmake_minimum_required(VERSION 3.13.1)
...
+  if(DEFINED TFM_ISOLATION_LEVEL)
+    set(TFM_ISOLATION_LEVEL_ARG -DTFM_ISOLATION_LEVEL=${TFM_ISOLATION_LEVEL})
+  endif()
+
+  if(DEFINED TFM_REGRESSION)
+    set(TFM_REGRESSION_ARG -DTEST_S=ON -DTEST_NS=ON)
+  endif()
+
+  if(DEFINED TFM_CMAKE_BUILD_TYPE)
+    set(TFM_CMAKE_BUILD_TYPE_ARG -DCMAKE_BUILD_TYPE=${TFM_CMAKE_BUILT_TYPE})
```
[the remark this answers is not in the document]
ioannisg: Fixed, thanks

ioannisg:
> @ioannisg Have we looked into the git-subtree utility? It's purpose is to do exactly what we're doing here. It ends up with the same results we have, just adding a few lines of metadata to the commits. But, it tracks downstream patches and makes this kind of thing a lot easier to manage.

I don't think we have consider this so far, but I agree we need to look at it. I 'd rather do this on the following upmerge, when we will be bringing in TF-M v.1.2.0

tejlmand (approved):

microbuilder:
When trying to perform a manual TF-M build with the new build system, both @agansari and myself get the following toolchain error:

```bash
$ cmake -S . -B build -G"Unix Makefiles" -DTFM_PLATFORM=mps2/an521 -DCMAKE_TOOLCHAIN_FILE=toolchain_GNUARM.cmake -DTFM_TEST_REPO_PATH=/Users/kevin/Linaro/tfm/buildsystem/tf-m-tests/
-- The C compiler identification is unknown
-- The ASM compiler identification is unknown
-- Found assembler: /bin/-gcc
CMake Error at CMakeLists.txt:10 (project):
  The CMAKE_C_COMPILER:

    /bin/-gcc

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:10 (project):
  The CMAKE_ASM_COMPILER:

    /bin/-gcc

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "ASM" or the CMake cache entry CMAKE_ASM_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Warning: Did not find file Compiler/-ASM
-- Configuring incomplete, errors occurred!
See also "/Users/kevin/Linaro/zephyr/trusted-firmware-m-zephyr/trusted-firmware-m/build/CMakeFiles/CMakeOutput.log".
See also "/Users/kevin/Linaro/zephyr/trusted-firmware-m-zephyr/trusted-firmware-m/build/CMakeFiles/CMakeError.log".
```

microbuilder:
This builds the TF-M samples with some variation of `$ west build -p -b mps2_an521_nonsecure zephyr/samples/tfm_integration/psa_level_1/ -t run`, but the `imgtool` module is missing, as per: https://ci.trustedfirmware.org/view/TF-M/job/tf-m-build-docs-nightly/lastStableBuild/artifact/trusted-firmware-m/build/install/doc/user_guide/html/docs/getting_started/tfm_sw_requirement.html#ubuntu-setup

The CI system will need to be updated to install `imgtool`, at which points builds work fine for me in QEMU on the AN521. Will test more systems shortly.

ioannisg:
@microbuilder  @agansari could I get you to re-visit the reviews on this upmerge PR?

anangl (approved):

microbuilder (approved):

CMakeLists.txt
```diff
@@ -65,33 +93,39 @@ function(trusted_firmware_build)
...
-                ${TFM_INTERFACE_SOURCE}/tfm_its_ipc_api.c
-                ${TFM_INTERFACE_SOURCE}/tfm_platform_ipc_api.c
-                ${TFM_INTERFACE_SOURCE}/tfm_ps_ipc_api.c
-                ${TFM_INTERFACE_SOURCE}/tfm_ns_interface.c
-    )
-    target_include_directories(tfm_ipc_psa_api PUBLIC ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include)
-    target_compile_definitions(tfm_ipc_psa_api PUBLIC TFM_PSA_API TFM_PARTITION_TEST_CORE_IPC)
-    target_link_libraries(tfm_ipc_psa_api PRIVATE zephyr_interface)
-  endif()
+  add_library(tfm_ipc_psa_api STATIC IMPORTED)
```
oyvindronningstad, line 112: We should probably change the name since both ipc and psa are configurable.

Maybe just "tfm_api" or "tfm_interface"?

init-git.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-TFM_GIT_SHA=TF-Mv1.1
-TFM_TESTS_GIT_SHA=TF-Mv1.1
+TFM_GIT_SHA=8bebd05745a8b27dccc6403f0215fa6e39de3bc1
+TFM_TESTS_GIT_SHA=652bb8a1502c95bfd1ee8e44ccd1126ce74ae41f
 MCUBOOT_GIT_SHA=v1.6.0
 MBED_CRYPTO_GIT_SHA=mbedcrypto-3.0.1
```
oyvindronningstad, line 6: IIRC, mbedcrypto is no longer used. Instead they use mbedtls 2.23.0 I think.

Update README.rst as well.

oyvindronningstad (commented):
Just a couple of things.

init-git.sh
ioannisg: Yeah, I've thought about this. The thing is, I am not currently removing the MBedCrypto directory, that's why I kept the entry here, and in the readme. 
On second thought, I'll start with removing the info from the init-git and readme, and then we can think of cleaning up the directories on the next upmerge.

CMakeLists.txt
ioannisg: tfm_api is good.
Will apply.

ioannisg: Done

init-git.sh
ioannisg: Applied.

oyvindronningstad (approved):

CMakeLists.txt
```diff
@@ -65,33 +93,37 @@ function(trusted_firmware_build)
...
+               -DTFM_PLATFORM=${TFM_BOARD}
+               ${TFM_CMAKE_BUILD_TYPE_ARG}
                ${TFM_BL2_ARG}
-               -DCOMPILER=${TFM_TOOLCHAIN}
-               -DGNUARM_PREFIX=${TFM_TOOLCHAIN_PREFIX}
-               -DGNUARM_PATH=${TFM_TOOLCHAIN_PATH}
+               ${TFM_IPC_ARG}
+               ${TFM_ISOLATION_LEVEL_ARG}
+               ${TFM_REGRESSION_ARG}
+               -DTFM_TEST_REPO_PATH=${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests
```
agansari, line 103: I would also add -DNS=OFF, I see no point in building the TFM NS image, we use Zephyr as NS image and Zephyr's cmake handles signing and image merge. Saves some build time.

ioannisg: Thanks @agansari , I would like to keep this out of the up-merge PR; this is sort of a change compared to the current state. But the proposal seems reasonable, so I will include this in a list of improvements for the TF-M module (I will open a dedicated ticket for that copy all TFM stakeholders)

PR #16, opened 2020-10-30
Commit mcuboot @ e8fe6cf

oyvindronningstad:
With most submodules.

This provides a 20-40 second speedup for pristine builds even on my fast connection.

timed with

    time wb samples/tfm_integration/tfm_ipc/ -p


oyvindronningstad:
I found a workaround: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/6703

PR #18, opened 2020-11-02
Allow flash layout to be set from cmake

oyvindronningstad:
Idea for setting flash and RAM layout via Cmake. Not really tested properly yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,9 @@ function(trusted_firmware_build)
     message(FATAL_ERROR "Unsupported ZEPHYR_TOOLCHAIN_VARIANT: ${ZEPHYR_TOOLCHAIN_VARIANT}")
   endif()
 
+  #math(EXPR flash_end "${PM_TFM_EXTRA_ADDRESS}+${PM_TFM_EXTRA_SIZE}")
+  set(flash_end 0x100000)
+
   include(ExternalProject)
   
   ExternalProject_Add(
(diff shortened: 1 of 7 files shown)


oyvindronningstad:
cc @ioannisg @tejlmand

d3zd3z:
Zephyr uses the device tree to get this information. How does setting it in cmake in tfm interact with the partition table and other information in the device tree within Zephyr? Have we considered using the device tree in the tfm build?

oyvindronningstad:
> 
> 
> Zephyr uses the device tree to get this information. How does setting it in cmake in tfm interact with the partition table and other information in the device tree within Zephyr? Have we considered using the device tree in the tfm build?

Device tree quickly becomes infeasible for this purpose, as we have discovered in nRF Connect SDK, where we have a script for partitioning the flash and RAM, leading us to abandon device tree for this specific purpose. Setting it from Cmake is the most flexible IMO since all values from DT/Kconfig are available in Cmake.

CMakeLists.txt
```diff
@@ -103,6 +106,15 @@ function(trusted_firmware_build)
                -DTFM_TEST_REPO_PATH=${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests
                -DTFM_TOOLCHAIN_PATH=${TFM_TOOLCHAIN_PATH}
                -DTFM_TOOLCHAIN_PREFIX=${TFM_TOOLCHAIN_PREFIX}
+
+               -DFLASH_AREA_START=$<TARGET_PROPERTY:partition_manager,PM_BL2_ADDRESS>
```
tejlmand, line 110: Where is `partition_manager` defined ?
Is there a corresponding Zephyr PR missing somewhere.

tejlmand (changes requested):
Initial feeling is that this looks overly complicated considering how TFM uses this.

- CMake file with default cache settings
- Settings are set as compile flags, which gives some verry long compile commands, and makes it harder for IDEs to lookup the values
- Zephyr needs to pass 8 variables to TFM build system.

Looking at: 
https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/partition
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/6edaaaaaa1a227cb1ccc9ef698a9d2e258a9bbd5/trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/CMakeLists.txt#L16-L18

https://github.com/zephyrproject-rtos/trusted-firmware-m/tree/master/trusted-firmware-m/platform/ext/target/musca_a/partition
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/6edaaaaaa1a227cb1ccc9ef698a9d2e258a9bbd5/trusted-firmware-m/platform/ext/target/musca_a/CMakeLists.txt#L13-L16

it seems that those flash layout files are placed in dedicated include folders.

So if Zephyr needs fine-grained control of the flash layout, then why not simply generate with the configured values, and then use that folder as include path ?

TFM could then keep their own set of headers for generic build, while allowing Zephyr to have it's own custom headers.

oyvindronningstad:
> TFM could then keep their own set of headers for generic build, while allowing Zephyr to have it's own custom headers.

I certainly like the idea, I had started thinking of generated files. How do you see the mechanism for deciding whether to override the default or not? Should it be possible to pass an include path to TFM via Cmake?

tejlmand:
> > TFM could then keep their own set of headers for generic build, while allowing Zephyr to have it's own custom headers.
> 
> I certainly like the idea, I had started thinking of generated files. How do you see the mechanism for deciding whether to override the default or not? Should it be possible to pass an include path to TFM via Cmake?

I have not at this point made extensive thought on such a design.

Setting the include path could be one way to do it.
That would allow a Zephyr board / TFM platform maintainer to introduce bigger changes in the platfrom files and have those changes outdside TFM repo, but it will also make it more complex / copy-paste more code just for a minor config change.

If the code itself is generic enough, and it is only a matter setting specific settings, then it could be considered to simply create a configuration header that could be passed to TFM.
- Zephyr build system creates `autoconf.h` which is used by all Zephyr sources for compilation.
- mbedTLS supports the use of `MBEDTLS_CONFIG_FILE`, and if not specified it uses a default file.

The benefit of setting the include path would be that it doesn't affect else than the specific platform, and can be handled directly in CMake.
Using a config header would require all platform headers from all vendors to support such a config header.

Let's look into possible solutions, and thanks a lot for raising this draft, I like the idea.

Maybe it could also be considered with this draft how to allow the TFM platform to be used easily by multiple different Zephyr boards using some SoC, as mentioned in this comment: https://github.com/zephyrproject-rtos/zephyr/pull/30145#issuecomment-730962083

oyvindronningstad:
Will instead be addressed by https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7301

PR #19, opened 2020-11-09
mcuboot: Commit v1.7.0-rc1

oyvindronningstad:
Retrieve mcuboot via west. See https://github.com/zephyrproject-rtos/zephyr/pull/30050.

This PR removes the existing committed mcuboot which is outdated.

~Commit MCUboot to avoid long download times on each build.~

~CMP0097 doesn't work on 3.16 and 3.17.
https://gitlab.kitware.com/cmake/cmake/-/issues/20579~

~Add a workaround (request non-existant submodule).
This prevents fetching almost 200 MB of data, improving
pristine build times.~

~This change was added upstream, but reverted because of unspecified build problems.
Add it here unless it gives problems, since it improves build times 40 seconds.~


oyvindronningstad:
This now has a PR upstream again: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7023

ioannisg:
FYI @agansari

ioannisg (approved):
Looks pretty straight forward to me.

ioannisg:
@oyvindronningstad would we need to manually upmerge the mcuboot in order to move forward from v1.7.0-RC1?

ioannisg:
@galak  @microbuilder could yo please add feedback?

oyvindronningstad:
> @oyvindronningstad would we need to manually upmerge the mcuboot in order to move forward from v1.7.0-RC1?

Yes

microbuilder:
This should stay in sync with the upstream TF-M repo in my opinion, but it looks like you're already working towards that: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7023

oyvindronningstad:
> 
> 
> This should stay in sync with the upstream TF-M repo in my opinion, but it looks like you're already working towards that: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7023
> 
> Edit: I +1'ed the upstream change request.

Note that this PR is now different from the ustream one. This one does not change the trusted-firmware-m directory precisely because we might not get the upstream change in for 1.2.

CMakeLists.txt
```diff
@@ -105,6 +105,7 @@ function(trusted_firmware_build)
                ${TFM_REGRESSION_ARG}
                ${TFM_PROFILE_ARG}
                -DTFM_TEST_REPO_PATH=${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests
+               -DMCUBOOT_PATH=${ZEPHYR_TFM_MODULE_DIR}/mcuboot
```
tejlmand, line 108: Why is it not possible to use:
```suggestion
               -DMCUBOOT_PATH=${ZEPHYR_MCUBOOT_MODULE_DIR}
```

and thus use the same mcuboot as Zephyr ?
And then completely remove mcuboot from this repository.

Are they not compatible ?

tejlmand (changes requested):

CMakeLists.txt
oyvindronningstad: They are, and when I tried it it works, they just aren't identical I think. So I think to keep TFM exactly as released, we don't want to do that yet. Though I'm not entirely convinced myself why we can't.

tejlmand: I understand that reason, but in that case, why not use `west` ?
That would remove the need to upmerge mcuboot in two places:
- This repo
- https://github.com/zephyrproject-rtos/mcuboot

Just add this to `west.yml`:

```
    - name: tfm_mcuboot
      repo-path: mcuboot
      path: modules/tee/tfm/mcuboot
      revision: 69344636bee1f18a6703dc5a8e7a0edfd8128d45
#      revision: 1.7.0-rc1
```
Note, the `69344636bee1f18a6703dc5a8e7a0edfd8128d45` SHA is part of this PR: https://github.com/zephyrproject-rtos/mcuboot/pull/39
https://github.com/zephyrproject-rtos/mcuboot/pull/39/commits/69344636bee1f18a6703dc5a8e7a0edfd8128d45

But if you ask @nvlsianpu, then maybe he can include the mcuboot tags in the upmerge, and you will be able to use `1.7.0-rc1` instead of the SHA.

That way we remove the need to upmerge mcuboot inside this repo, but still use a different version for TFM, than what Zephyr is using.
And there is only one place to keep mcuboot in sync with upstream.

nvlsianpu: > But if you ask @nvlsianpu, then maybe he can include the mcuboot tags in the upmerge, and you will be able to use 1.7.0-rc1 instead of the SHA

I can send 1.7.0-rc1 tag to zephyr's mcuboot fork If needed.

oyvindronningstad: Changed to use west now. See if you like it better

microbuilder: I'm OK with this approach, but we should use the exact MCUBoot commit they are using in upstream TF-M at the point that we update the Zephyr TF-M module here, just to avoid any subtle issues or complications.

tejlmand: @microbuilder that would be the optimal, and if this guide is followed:
https://docs.zephyrproject.org/latest/guides/modules.html#allowed-practices
>  Upstream changes brought in by performing a Git merge of the intended upstream branch (e.g. main branch, latest release branch, etc.) submitting the result in pull request against the module main branch, and merging the pull request using the Create a merge commit operation. This approach is applicable to modules with an upstream project Git repository. The main advantages of this approach is that the upstream repository history (that is, the original commit SHAs) is preserved in the module repository. 

then the upstream SHAs should be available.

tejlmand: > Changed to use west now. See if you like it better

You need a commit with:
```
git rm -r mcuboot
```
for this to work.

tejlmand: > @microbuilder that would be the optimal, and if this guide is followed:

Just verified, https://github.com/zephyrproject-rtos/mcuboot/pull/39, does have all the upstream SHAs.

oyvindronningstad: Should I change the west entry to a sha, or should we bring in the tag?

oyvindronningstad: > You need a commit with:
> 
> ```
> git rm -r mcuboot
> ```
> 
> for this to work.

Done

tejlmand: > > But if you ask @nvlsianpu, then maybe he can include the mcuboot tags in the upmerge, and you will be able to use 1.7.0-rc1 instead of the SHA
> 
> I can send 1.7.0-rc1 tag to zephyr's mcuboot fork If needed.

@nvlsianpu that would be highly appreciated :smile:

tejlmand (approved):
Much nicer.

oyvindronningstad:
> 
> 
> This should stay in sync with the upstream TF-M repo in my opinion, but it looks like you're already working towards that: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7023
> 
> Edit: I +1'ed the upstream change request.

Thanks, but now this change is entirely outside of the upstream repo. Can you take another look with this in mind?

oyvindronningstad:
@microbuilder @galak Please review

microbuilder:
@d3zd3z Care to give your feedback on this as well?

.gitignore
```diff
@@ -0,0 +1 @@
+/mcuboot
```
galak, line 1: this seems a little weird.

galak (changes requested):
I don't think it makes sense to assume that the mcuboot from zephyr west will be in sync with TFM.

microbuilder:
> I don't think it makes sense to assume that the mcuboot from zephyr west will be in sync with TFM.

I sort of dislike the assumption here as well, and the additional maintenance burden to keep Zephyr's MCUBoot requirements in sync with TF-M's.

oyvindronningstad:
> 
> 
> > I don't think it makes sense to assume that the mcuboot from zephyr west will be in sync with TFM.
> 
> I sort of dislike the assumption here as well, and the additional maintenance burden to keep Zephyr's MCUBoot requirements in sync with TF-M's.

@galak @microbuilder This solution was agreed on a while ago in this thread: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/19#discussion_r526176205

It seems there's no optimal way to do this, so I suggest that we merge this. We can always redo it later.

I really want to stop wasting a whole minute every time I build TFM.

galak:
> @galak @microbuilder This solution was agreed on a while ago in this thread: [#19 (comment)](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/19#discussion_r526176205)
> 
> It seems there's no optimal way to do this, so I suggest that we merge this. We can always redo it later.
> 
> I really want to stop wasting a whole minute every time I build TFM.

I suggest looking for other solutions to the issue.  I don't agree with this approach of coupling the mcuboot. @d3zd3z do you mind commenting as well.

d3zd3z (changes requested):
I'm a bit uncomfortable having TFM use the mcuboot out of the Zephyr repository rather than the upstream. The fork that is in Zephyr has the purpose of holding onto Zephyr-specific support before it can be upstreamed. There is no testing for whether that repo works in any non-zephyr configuration, including TF-M. In fact, it is likely to lag behind the upstream, where TF-M-specific fixes will end up.

Upstream TF-M will be removing the mcuboot code from the repo, and referencing upstream. I think we should mirror that. If we have timing issues with builds, maybe that can be addressed another way. Pointing at the wrong repo is not the right answer, though.

d3zd3z:
BTW, if the concern is pulling things with west that aren't needed, that could be fixed by having a project file that doesn't bring in dependencies we don't use. That could be mcuboot, and possibly other repos.

oyvindronningstad:
@d3zd3z the commit being pulled for MCUboot is the exact same one used by upstream TFM, even if it's taken from Zephyr's fork.

oyvindronningstad:
> 
> 
> > @galak @microbuilder This solution was agreed on a while ago in this thread: [#19 (comment)](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/19#discussion_r526176205)
> > It seems there's no optimal way to do this, so I suggest that we merge this. We can always redo it later.
> > I really want to stop wasting a whole minute every time I build TFM.
> 
> I suggest looking for other solutions to the issue. I don't agree with this approach of coupling the mcuboot. @d3zd3z do you mind commenting as well.

I already completely reworked this PR twice. I can do it a third time, but please give me more to work with than "I don't agree".

I don't care what the solution is, I just want to get something in. This is bothering me every day working with TFM.

galak:
> > 
> 
> I already completely reworked this PR twice. I can do it a third time, but please give me more to work with than "I don't agree".
> 
> I don't care what the solution is, I just want to get something in. This is bothering me every day working with TFM.

I understand.  Based on your comment to @d3zd3z I think maybe we don't understand how this works.

Now that I see what's going on, I've got no issue.  Sorry about the noise.

galak (approved):

d3zd3z (approved):
I misunderstood the branch being used. I'm ok with using this, then.

galak:
@oyvindronningstad what is the whole minute you are referring to associated with?

galak (changes requested):
I'd like to see if we can have the west.yml just live in the zephyr/trusted-firmware-m repo instead of having this being set in the zephyr/west.yml.

oyvindronningstad:
> @oyvindronningstad what is the whole minute you are referring to associated with?

Mcuboot is being downloaded, with all its submodules. This is approx 200MB, taking around 1 min on my connection.

oyvindronningstad:
> I misunderstood the branch being used. I'm ok with using this, then.

Glad we cleared it up 👍

galak:
> Mcuboot is being downloaded, with all its submodules. This is approx 200MB, taking around 1 min on my connection.

If I understand tfm's cmake is downloading mcuboot in the case you are describing for every new build you do.  Is that correct?

We just need a solution to point to a "pre-cloned" version of mcuboot, so that every tfm build doesn't checkout a new version of mcuboot.  We could do this by just updating the files in `zephyrproject-rtos/trusted-firmware-m` for mcuboot.

oyvindronningstad:
> We just need a solution to point to a "pre-cloned" version of mcuboot, so that every tfm build doesn't checkout a new version of mcuboot. We could do this by just updating the files in `zephyrproject-rtos/trusted-firmware-m` for mcuboot.

Exactly, and that was my 1st rewrite. It was rejected in https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/19#discussion_r526176205.

I suggest we merge this. The MCUboot situation is not figured out yet, so we will have plenty of opportunity to change things later.

galak:
@oyvindronningstad can we tweak this to introduce a `west.yml` in the `trusted-firmware-m` repo and use `import` in `zephyr/west.yml`

so `trusted-firmware-m/west.yml` would look like:

```
manifest:
  projects:
    - name: tfm-mcuboot
      url: https://github.com/zephyrproject-rtos/mcuboot
      revision: 69344636bee1f18a6703dc5a8e7a0edfd8128d45 
```

galak:
Also, I think if we are going down this path, we should do it for mbed-crypto as well.  I can setup the zephyrproject mirror if you are in agreement.

oyvindronningstad:
> @oyvindronningstad can we tweak this to introduce a `west.yml` in the `trusted-firmware-m` repo and use `import` in `zephyr/west.yml`

That sounds like a good plan. I'll add it.

>Also, I think if we are going down this path, we should do it for mbed-crypto as well. I can setup the zephyrproject mirror if you are in agreement.

Sounds good as well. Note (in case you are nit aware) that even though they call it mbedcrypto they are using mbedtls. Also: they modify the repo with a number of patches. These modifications should be committed to the repo mirror.

Thanks!

galak:
Also, I pushed the 'v1.7.0', `1.7.0-rc1` and `v1.7.0-rc2` tags to the zephyr mcuboot repo so we can use those now.

galak:
> Sounds good as well. Note (in case you are not aware) that even though they call it mbedcrypto they are using mbedtls. Also: they modify the repo with a number of patches. These modifications should be committed to the repo mirror.

I was not aware.  Looking at the tfm buildsystem, I think we need to tweak things so that tfm does the patch'ing.

We can deal with mbedtls/mbed-crypto after we get mcuboot working.

ioannisg:
@galak could you please take a final look?

west.yml
```diff
@@ -0,0 +1,15 @@
...
+#
+# See the west documentation for more information:
+#
+# https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/west/index.html
+
+manifest:
+  projects:
+    - name: tfm-mcuboot
+      url: https://github.com/zephyrproject-rtos/mcuboot
+      path: modules/tee/tfm/mcuboot
```
galak, line 14: Is it possible that this can be `path: modules/tee/tfm-mcuboot`.  The reason is because of how `west` works have this be a different path from the existing `mcuboot` path makes life a bit easier if one goes back in history w/a checkout.

.gitignore
galak: Is this correct?

galak (changes requested):
In general looks good:

Just the 2 questions:
* the .gitignore with /mcuboot seems weird, and maybe wrong or not needed
* can we tweak the path of where mcuboot gets put by west.

west.yml
ioannisg: Yes, I think this is reasonable

oyvindronningstad: Done

.gitignore
oyvindronningstad: It was needed since west checked out to here. Removed now that mcuboot has moved to ../tfm-mcuboot

README.rst
```diff
@@ -24,14 +24,10 @@ TF-M Tests:
...
-    commit: 50d24a57516f558dac72bef634723b60c5cfb46b
-    license: Apache 2.0
-
 psa-arch-tests:
     repo: https://github.com/ARM-software/psa-arch-tests
     tag: v20.03_API1.0
     commit: 3d986e72d9e084bb45763849699718f474fb10ba
     license: Apache 2.0
+
+See also west.yml for more dependencies.
```
ioannisg, line 33: I wonder if you could add a more detailed comment here, to explain what happens with MCUboot, but it is not a must-do for this PR.

galak (approved):

galak:
Guess we need a zephyr side PR now.

PR #20, opened 2020-11-09
Cmake cleanup

oyvindronningstad:
clenaup and improvments to make later changes easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,11 +65,15 @@ function(trusted_firmware_build)
 
   set(VENEERS_FILE ${TFM_BINARY_DIR}/secure_fw/s_veneers.o)
   set(PSA_API_NS_PATH ${TFM_BINARY_DIR}/interface/libpsa_api_ns.a)
+  set(TFM_GENERATED_INCLUDES ${TFM_BINARY_DIR}/generated/interface/include)
+
+  set(BL2_HEX_FILE ${CMAKE_BINARY_DIR}/tfm/bin/bl2.hex)
 
   set(BUILD_BYPRODUCTS
     ${VENEERS_FILE}
     ${PSA_API_NS_PATH}
     ${BL2_HEX_FILE}
+    ${TFM_GENERATED_INCLUDES}/psa_manifest/sid.h
     )
 
   # Get the toolchain variant
@@ -88,7 +92,7 @@ function(trusted_firmware_build)
   endif()
 
   include(ExternalProject)
-  
+
   ExternalProject_Add(
     tfm
     SOURCE_DIR ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m
@@ -110,23 +114,27 @@ function(trusted_firmware_build)
     BUILD_BYPRODUCTS ${BUILD_BYPRODUCTS}
   )
 
-  add_library(tfm_api STATIC IMPORTED)
-  set_target_properties(tfm_api PROPERTIES
-    IMPORTED_LOCATION ${PSA_API_NS_PATH}
+  add_library(tfm_api
+    ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/app/os_wrapper_cmsis_rtos_v2.c
   )
 
-  add_library(veneer_lib STATIC IMPORTED)
-  set_target_properties(veneer_lib PROPERTIES
-    IMPORTED_LOCATION ${VENEERS_FILE}
+  target_include_directories(tfm_api
+    PRIVATE
+    ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/CMSIS/RTOS2/Include
+    PUBLIC
+    ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
+    INTERFACE
+    ${TFM_GENERATED_INCLUDES}
+  )
+
+  target_link_libraries(tfm_api
+    PUBLIC
+    zephyr_interface
+    INTERFACE
+    ${PSA_API_NS_PATH}
+    ${VENEERS_FILE}
+    $<TARGET_FILE:tfm_api>
   )
 
   add_dependencies(tfm_api tfm)
-  file(MAKE_DIRECTORY ${TFM_BINARY_DIR}/generated/interface/include)
-  target_include_directories(tfm_api INTERFACE
-    ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
-    ${TFM_BINARY_DIR}/generated/interface/include
-  )
-
-  target_link_libraries(tfm_api INTERFACE zephyr_interface veneer_lib)
-
 endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-# Adds trusted-firmware-m as an external project, and provides output values
-# that are required to link against the secure TF-M binary.
+# Adds trusted-firmware-m as an external project.
+# Also creates a target called 'tfm_api'
+# which can be linked into the app.
 #
 # When called from a Zephyr module, the following input values can be provided
 # to configure the TF-M build:
@@ -18,12 +19,6 @@
 # REGRESSION: Boolean if TF-M build includes building the TF-M regression tests
 # BL2: Boolean if the TF-M build uses MCUboot.
 #
-# The following output values can also be used:
-#
-# OUT_VENEERS_FILE: The path and filename of the veneer library to link against,
-#                   which identifies where the veneer functions are in memory.
-#                   Should be added via "target_link_libraries"
-#
 # Example usage:
 #
 # trusted_firmware_build(BINARY_DIR ${CMAKE_BINARY_DIR}/tfm
@@ -33,10 +28,10 @@
 #                        ISOLATION_LEVEL 2
 #                        REGRESSION
 #                        BL2 True
-#                        BUILD_PROFILE profile_small
+#                        BUILD_PROFILE profile_small)
 function(trusted_firmware_build)
   set(options IPC REGRESSION)
-  set(oneValueArgs BINARY_DIR BOARD OUT_INCLUDE_PATH BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
+  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
   cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "" ${ARGN})
 
   if(DEFINED TFM_BL2)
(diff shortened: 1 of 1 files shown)


oyvindronningstad:
cc @ioannisg @tejlmand @anangl

CMakeLists.txt
```diff
@@ -62,6 +62,12 @@ function(trusted_firmware_build)
   set(VENEERS_FILE ${TFM_BINARY_DIR}/secure_fw/s_veneers.o)
   set(PSA_API_NS_PATH ${TFM_BINARY_DIR}/interface/libpsa_api_ns.a)
 
+  set(BL2_HEX_FILE ${CMAKE_BINARY_DIR}/tfm/bin/bl2.hex)
+  set_property(GLOBAL PROPERTY
+    bl2_PM_HEX_FILE
```
ioannisg, line 67: @oyvindronningstad what is bl2_PM_HEX_FILE needed for?

ioannisg:
@anangl could you take a look?

ioannisg:
I rebased that myself @oyvindronningstad , pls address the comment I left.
@microbuilder I wonder if you can take a look at this cleanup PR.

CMakeLists.txt
oyvindronningstad: It's NCS-specific. We can probably add it from there instead, and remove it from here. I meant to remove it, just forgot.

ioannisg: Yes, please, clean this PR up so it is not specific to the Nordic SDK :)

CMakeLists.txt
```diff
@@ -102,27 +107,41 @@ function(trusted_firmware_build)
...
   )
 
-  # IPC mode source dependencies
-  add_library(tfm_api STATIC IMPORTED)
-  set_target_properties(tfm_api PROPERTIES
+  add_library(tfm_api
+    ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/app/os_wrapper_cmsis_rtos_v2.c
+  )
+
+  add_library(interface_lib STATIC IMPORTED)
```
tejlmand, line 117: `interface_lib` seems to be a too generic name.

Who needs to link to this lib, and why ?
Based on this `PSA_API_NS_PATH` I think the name could be improved.

However, from what I see what you want is probably to ensure that targets depending on `tfm_api` also links to this librarym that is: `target_link_libraries(<target>... tfm_api)` also links with `PSA_API_NS_PATH`.

For this, you should not use imported lib, as that means the build system expects that lib is always present, but in this case that is not true. You are using a generated lib.

So you probably should do something like here:
https://github.com/nrfconnect/sdk-nrfxlib/blob/master/nrf_security/cmake/combine_archives.cmake#L28-L29

Form what I can judged, something like:
```
target_link_libraries(tfm_api INTERFACE ${PSA_API_NS_PATH})
add_dependencies(tfm_api tfm)   # <--- This you already have in line 138
```

tejlmand (changes requested):
A suggestion, and some questions.

CMakeLists.txt
```diff
@@ -102,27 +107,41 @@ function(trusted_firmware_build)
...
   file(MAKE_DIRECTORY ${TFM_BINARY_DIR}/generated/interface/include)
-  target_include_directories(tfm_api INTERFACE
+
+  target_include_directories(tfm_api
+    PRIVATE
+    ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/CMSIS/RTOS2/Include
+    PUBLIC
     ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
+    INTERFACE
     ${TFM_BINARY_DIR}/generated/interface/include
```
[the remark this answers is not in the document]
oyvindronningstad: Turns out it was safe to remove this. Thanks for pointing it out.

CMakeLists.txt
oyvindronningstad: I separated some of tfm_api into this:

    add_library(tfm_api_helpers
      ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/app/os_wrapper_cmsis_rtos_v2.c
      ${ZEPHYR_TFM_MODULE_DIR}/src/zephyr_tfm_log.c
    )

    target_include_directories(tfm_api_helpers
      PRIVATE
      ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/CMSIS/RTOS2/Include
      PUBLIC
      ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
    )

    target_link_libraries(tfm_api_helpers PUBLIC
      zephyr_interface
    )

I need tfm_api_helpers to be linked into the generated libs (like `${PSA_API_NS_PATH}`). How do I do this?

oyvindronningstad: I've tried to achieve it in many different ways, but I cannot find a way where the generated libs find the symbols from tfm_api_helpers, except to create imported libs like before.

CMakeLists.txt
```diff
@@ -32,10 +28,10 @@
 #                        IPC
 #                        ISOLATION_LEVEL 2
 #                        REGRESSION
-#                        BL2 True
+#                        BL2 True)
 function(trusted_firmware_build)
   set(options IPC REGRESSION)
-  set(oneValueArgs BINARY_DIR BOARD OUT_INCLUDE_PATH BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE)
+  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE)
```
ioannisg, line 34: Beware, here, of #29 which adds the BUILD_PROFILE in the list of arguments.

ioannisg:
This needs rebase now @oyvindronningstad , since #29 was merged.

CMakeLists.txt
oyvindronningstad: Done

CMakeLists.txt
oyvindronningstad: Got help offline, done.

CMakeLists.txt
oyvindronningstad: Rebased

CMakeLists.txt
```diff
@@ -109,27 +112,30 @@ function(trusted_firmware_build)
                -DTFM_TOOLCHAIN_PREFIX=${TFM_TOOLCHAIN_PREFIX}
     BUILD_ALWAYS True
     USES_TERMINAL_BUILD True
-    BUILD_BYPRODUCTS ${VENEERS_FILE} ${PSA_API_NS_PATH}
```
tejlmand, line 112: Why are we running the install step also ?

Why not disable the install stage with
```
 INSTALL_COMMAND ""
```
because you are already using:
```
${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
```
in line 129.

So you only need:
```
${TFM_BINARY_DIR}/generated/interface/include/psa_manifest
```

microbuilder:
This lgtm but I'll wait for @tejlmand to accept the changes he requested.

CMakeLists.txt
```diff
@@ -69,10 +64,18 @@ function(trusted_firmware_build)
 
 
   set(VENEERS_FILE ${TFM_BINARY_DIR}/secure_fw/s_veneers.o)
-  set(${TFM_OUT_VENEERS_FILE} ${VENEERS_FILE} PARENT_SCOPE)
-  set(${TFM_OUT_INCLUDE_PATH} ${TFM_BINARY_DIR}/install/export/tfm/include PARENT_SCOPE)
-
   set(PSA_API_NS_PATH ${TFM_BINARY_DIR}/interface/libpsa_api_ns.a)
+  set(TFM_EXPORTED_INCLUDES ${TFM_BINARY_DIR}/install/export/tfm/include)
+  set(SID_H_FILE_PATH ${TFM_EXPORTED_INCLUDES}/psa_manifest/sid.h)
```
tejlmand, line 69: Can be changed to:
```suggestion
${TFM_BINARY_DIR}/generated/interface/include/psa_manifest/sid.h
```
if we want to disable the install step.

CMakeLists.txt
```diff
@@ -109,27 +112,30 @@ function(trusted_firmware_build)
...
-  file(MAKE_DIRECTORY ${TFM_BINARY_DIR}/generated/interface/include)
-  target_include_directories(tfm_api INTERFACE
+  target_include_directories(tfm_api
+    PRIVATE
+    ${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests/CMSIS/RTOS2/Include
+    PUBLIC
     ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include
-    ${TFM_BINARY_DIR}/generated/interface/include
+    INTERFACE
+    ${TFM_BINARY_DIR}/install/export/tfm/include
```
tejlmand, line 128: you are already using: `${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include` above as PUBLIC, which has the same content (+more) as the `export/tfm/include`.

Only content in `export/tfm/include` you might need is the generated `sid.h` which is in:
```
${TFM_BINARY_DIR}/generated/interface/include/psa_manifest
```

tejlmand (changes requested):
Much better.

Think we can disable the install step, let's discuss that.

CMakeLists.txt
oyvindronningstad: The install step is needed for the signing commands. They access the bin files via there. That can probably be fixed later, but is IMO out of scope of this PR.

oyvindronningstad: I changed the path anyways.

CMakeLists.txt
oyvindronningstad: I need `${TFM_BINARY_DIR}/generated/interface/include/psa_manifest`
Would you prefer I remove `${TFM_BINARY_DIR}/generated/interface/include`?

CMakeLists.txt
oyvindronningstad: Will do

CMakeLists.txt
tejlmand: > The install step is needed for the signing commands. They access the bin files via there. 
Thanks for the explanation.
Could you make a comment on the ExternalProject, why the install step is needed :)

CMakeLists.txt
tejlmand: >  the convention is to include sid.h as #include "psa_manifest/sid.h"

Good to hear.

> Would you prefer I remove ${TFM_BINARY_DIR}/generated/interface/include

You need it, for users of tfm to be able to include `sid.h`.
But using the generated path instead of the install path is nicer as it will make it easier to remove the install step in the future.

oyvindronningstad: Sorry, I meant would you prefer I remove `${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include`?

CMakeLists.txt
```diff
@@ -69,10 +64,18 @@ function(trusted_firmware_build)
...
+  set(TFM_GENERATED_INCLUDES ${TFM_BINARY_DIR}/generated/interface/include)
+  set(SID_H_FILE_PATH ${TFM_GENERATED_INCLUDES}/psa_manifest/sid.h)
+
+  set(BL2_HEX_FILE ${CMAKE_BINARY_DIR}/tfm/bin/bl2.hex)
+
+  set(BUILD_BYPRODUCTS
+    ${VENEERS_FILE}
+    ${PSA_API_NS_PATH}
+    ${BL2_HEX_FILE}
+    ${SID_H_FILE_PATH}
```
tejlmand, line 77: Minor.
`SID_H_FILE_PATH` is only used here, so there is not a big value in creating a variable for it.

tejlmand (approved):
Approved. Thanks for this.

One minor comment, but PR is also approved as-is.

CMakeLists.txt
tejlmand: as you were adding `export/tfm/include` in earlier revision, then I think you need either `export/tfm/include` or `${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include`, right ?

And in that case, I prefer the current version with: `${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m/interface/include`

CMakeLists.txt
oyvindronningstad: Done

ioannisg (approved):

microbuilder (approved):

PR #21, opened 2020-11-09
Add kconfig and build TF-M based on it

oyvindronningstad:
Allow specifying options in Kconfig instead of via Cmake.

This also allows TF-M to be enabled in subsystems, since the
cmake function isn't available in subsystems, but the config is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,3 +138,35 @@ function(trusted_firmware_build)
 
   add_dependencies(tfm_api tfm)
 endfunction()
+
+if (CONFIG_BUILD_WITH_TFM)
+  if (CONFIG_TFM_IPC)
+    set(TFM_IPC_ARG IPC)
+  endif()
+  if (CONFIG_TFM_REGRESSION)
+    set(TFM_REGRESSION_ARG REGRESSION)
+  endif()
+  if (CONFIG_TFM_BL2_TRUE)
+    set(TFM_BL2_ARG BL2 True)
+  elseif (CONFIG_TFM_BL2_FALSE)
+    set(TFM_BL2_ARG BL2 False)
+  endif()
+  if (CONFIG_TFM_ISOLATION_LEVEL)
+    set(TFM_ISOLATION_LEVEL_ARG ISOLATION_LEVEL ${CONFIG_TFM_ISOLATION_LEVEL})
+  endif()
+  if (CONFIG_TFM_PROFILE)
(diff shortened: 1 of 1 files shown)


oyvindronningstad:
cc @tejlmand @ioannisg @anangl

oyvindronningstad:
@tejlmand Can you look at this first, since it has a lot of bearing on the direction of the NCS integration?

Kconfig
```diff
@@ -0,0 +1,16 @@
+menuconfig TFM
```
ioannisg, line 1: @oyvindronningstad could this be moved to the Zephyr side instead of the module?
E.g. in modules/...../tfm, or, maybe a subsystem if module directory is not good enough

ioannisg: Please move this into the main zephyr tree

zephyr/module.yml
```diff
@@ -1,2 +1,3 @@
 build:
   cmake: .
+  kconfig: Kconfig
```
ioannisg, line 3: and i guess this needs to be reverted

CMakeLists.txt
```diff
@@ -129,3 +129,29 @@ function(trusted_firmware_build)
   target_link_libraries(tfm_api INTERFACE zephyr_interface veneer_lib)
 
 endfunction()
+
+if (CONFIG_TFM)
```
ioannisg, line 133: How about the option config BUILD_WITH_TFM? Does this new option sort of overlap?

Kconfig
oyvindronningstad: Done

CMakeLists.txt
oyvindronningstad: Done

zephyr/module.yml
oyvindronningstad: Done

ioannisg (changes requested):
Please update commit title to Zephyr compliant.
E.g. trusted-firmware-m: build tf-m based on new Kconfig options

ioannisg (commented):
@microbuilder @tejlmand @anangl pls take a look if possible.
To me this is a good improvement, because the call to trusted_firmware_build does not need to be replicated in the samples any more.

ioannisg (approved):

microbuilder (approved):

anangl (approved):

agansari (approved):

PR #22, opened 2020-11-09
Native TFM ns app

oyvindronningstad:
Allow building and flashing the native TFM NS app instead of the Zephyr app, e.g. to run the vanilla regression tests as they appear in upstream.

Based on #20 and #21


microbuilder:
Cross reference proposed TFM KConfig changes on the Zephyr side here to see where the right place to define these is: https://github.com/zephyrproject-rtos/zephyr/pull/30013

ioannisg:
Needs rebase (At least #20 is merged already)

oyvindronningstad:
Superseeded by #34

PR #25, opened 2020-11-09
psa-arch-tests: Update PSA arch tests repo

oyvindronningstad:
repo: https://github.com/ARM-software/psa-arch-tests
commit: 5c0399f2ca8ab921d5b3f9643a9a2b57b47191ec


oyvindronningstad:
We will probably do this with west if we do it at all.

PR #28, opened 2020-11-11
Trusted Firmware M: Bump to version TF-M 1.2-RC2

ioannisg:
Upmerge to TFM v1.2.0 RC2
This is done to get support for nRF5340 DK, since nRF5340 PDK has been deprecated.

- trusted-firmware-m and tf-m-tests are update to v1.2.0 RC2
- minor updates in CMakeLists.txt
- two patches for the STM platform are carried over


ioannisg:
@agansari @anangl @oyvindronningstad  FYI

CMakeLists.txt
```diff
@@ -74,13 +74,11 @@ function(trusted_firmware_build)
...
+    set(TFM_TOOLCHAIN_FILE "trusted-firmware-m/toolchain_GNUARM.cmake")
     set(TFM_TOOLCHAIN_PREFIX "arm-zephyr-eabi")
-    set(TFM_TOOLCHAIN_PATH ${ZEPHYR_SDK_INSTALL_DIR}/arm-zephyr-eabi)
+    set(TFM_TOOLCHAIN_PATH ${ZEPHYR_SDK_INSTALL_DIR}/arm-zephyr-eabi/bin)
   elseif(${ZEPHYR_TOOLCHAIN_VARIANT} STREQUAL "gnuarmemb")
-    set(TFM_TOOLCHAIN "GNUARM")
-    set(CMAKE_TOOLCHAIN_FILE "trusted-firmware-m/toolchain_GNUARM.cmake")
+    set(TFM_TOOLCHAIN_FILE "trusted-firmware-m/toolchain_GNUARM.cmake")
     set(TFM_TOOLCHAIN_PREFIX "arm-none-eabi")
     set(TFM_TOOLCHAIN_PATH ${GNUARMEMB_TOOLCHAIN_PATH})
```
microbuilder, line 83: This should have `/bin` appended when not using the Zephyr SDK
```suggestion
    set(TFM_TOOLCHAIN_PATH ${GNUARMEMB_TOOLCHAIN_PATH}/bin)
```

microbuilder (changes requested):

CMakeLists.txt
ioannisg: Updated now @microbuilder , thanks.

oyvindronningstad (approved):

galak:
Just curious if the 2 changes for stm32l5 have been sent to upstream TFM?

microbuilder (approved):
Tested on `mps2_an521_nonsecure` (with QEMU) and `lpcxpresso55s69_ns` (HW), using https://github.com/zephyrproject-rtos/zephyr/pull/29956

ioannisg:
> Just curious if the 2 changes for stm32l5 have been sent to upstream TFM?

@galak, thanks, let me give you an update here:

I am planning to assist myself in upstreaming these patches to TFM (or, rather, make TFM a bit configurable so we do not need the patches any more). They are not significant patches, they exist just because of paths and binary names being hardcoded in the STM scripts.
- d8641e9 modifies the name and path of the NS signed binary when we build NS using Zephyr.
- 9e9f64b passes the compiler name/path the the STM script in TFM which is called via Zephyr.

As I said, I am planning to keep these two patches donwstream until they are sort of merged upstream. They are not controversial at all.

Good news is that with this upmerge we got rid of two other downstream patches
- some functional STM platform fixes
- the hack to pass the toolchain path to TFM (was fixed in tFM)

tejlmand (approved):
lgtm

anangl (approved):

tejlmand (changes requested):
Should also update here to 3.15:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/862316cf8148e539684f49c90ef9f114a9bda461/CMakeLists.txt#L6

There are a couple of more CMake files with 3.13, but those seem to come from upstream repos.

agansari (approved):

tejlmand (approved):

ioannisg:
Pushed a downstream bug for nRF5340 DK that did not make it to RC2 but the patch has been submitted to the upstream project, has been reviewed and will make it to RC3. So it will be removed in the next upmerge. Only nRF5340 DK specific (does not affect anything else).

https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/6981

PR #29, opened 2020-11-13
CMakeLists.txt: add BUILD_PROFILE

agansari:
Allows boards to use the non-default build profiles.
Used by LPC55S69 to build with 'profile_medium'.

Required by: zephyrproject-rtos/zephyr#30013

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,9 +33,10 @@
 #                        ISOLATION_LEVEL 2
 #                        REGRESSION
 #                        BL2 True
+#                        BUILD_PROFILE profile_small
 function(trusted_firmware_build)
   set(options IPC REGRESSION)
-  set(oneValueArgs BINARY_DIR BOARD OUT_INCLUDE_PATH BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


agansari:
@microbuilder @ioannisg @MaureenHelm @galak @d3zd3z

galak (approved):

ioannisg:
@oyvindronningstad FYI

ioannisg (approved):

ioannisg:
@agansari is there a Zephyr PR that runs CI with this update?

agansari:
This is intended for https://github.com/zephyrproject-rtos/zephyr/pull/30013, but should run without it, the parameter is optional, any CI will do.

ioannisg:
> This is intended for [zephyrproject-rtos/zephyr#30013](https://github.com/zephyrproject-rtos/zephyr/pull/30013), but should run without it, the parameter is optional, any CI will do.
> Added Zephyr PR do description.

@agansari in order to get this PR merged we need a .west.yml update with TF-M to point to _this_ PR instead of the current SHA.
Since this PR is a generic improvement, I think you should open a Zepyr PR with just the manifest update, and see whether CI remains green 👍

ioannisg:
Alternatively we could include the manifest update in #30013 but I don't see why we cannot do this separately (and it will be faster).

agansari:
@ioannisg will create a separate PR with west.yml, as i forgot to do that in  zephyrproject-rtos/zephyr#30013; it is faster.

microbuilder (approved):

PR #30, opened 2020-11-19
CMakeLists.txt: rework TF-M platform directory selection

ioannisg:
Rework how the platform directory is selected
based on Kconfig.

Signed-off-by: Ioannis Glaropoulos <Ioannis.Glaropoulos@nordicsemi.no>

Link to the relevant discussion: https://github.com/zephyrproject-rtos/zephyr/pull/30062#discussion_r524413356

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,25 @@ function(trusted_firmware_build)
   set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
   cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "" ${ARGN})
 
+  # Determine TF-M platform target
+  if(CONFIG_TFM_BOARD_LPCXPRESSO55S69)
+    set(TFM_PLATFORM_DIR "nxp/lpcxpresso55s69")
+  elseif(CONFIG_TFM_BOARD_MPS2_AN521)
+    set(TFM_PLATFORM_DIR "mps2/an521")
+  elseif(CONFIG_TFM_BOARD_MUSCA_B1)
+    set(TFM_PLATFORM_DIR "musca_b1")
+  elseif(CONFIG_TFM_BOARD_NRF5340DK_NRF5340)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf5340dk_nrf5340_cpuapp")
+  elseif(CONFIG_TFM_BOARD_NRF5340PDK_NRF5340)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf5340pdk_nrf5340_cpuapp")
+  elseif(CONFIG_TFM_BOARD_NRF9160DK_NRF9160)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf9160dk_nrf9160")
+  elseif(CONFIG_TFM_BOARD_BOARD_NUCLEO_L552ZE_Q)
+    set(TFM_PLATFORM_DIR "stm/nucleo_l552ze_q")
+  else()
+    message(FATAL_ERROR "Unsupported TFM BOARD")
+  endif()
+
   if(DEFINED TFM_BL2)
     set(TFM_BL2_ARG "-DBL2=${TFM_BL2}")
   endif()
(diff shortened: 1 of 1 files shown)


ioannisg:
CC @erwango

agansari (commented):
Think you should point to the discussion in the PR description.

galak (approved):

agansari (approved):

CMakeLists.txt
```diff
@@ -98,7 +117,7 @@ function(trusted_firmware_build)
     SOURCE_DIR ${ZEPHYR_TFM_MODULE_DIR}/trusted-firmware-m
     BINARY_DIR ${TFM_BINARY_DIR}
     CMAKE_ARGS -DTFM_TOOLCHAIN_FILE=${ZEPHYR_TFM_MODULE_DIR}/${TFM_TOOLCHAIN_FILE}
-               -DTFM_PLATFORM=${TFM_BOARD}
+               -DTFM_PLATFORM=${TFM_PLATFORM_DIR}
```
galak, line 120: Can't this just be `-DTFM_PLATFORM=${CONFIG_TFM_BOARD}

galak (changes requested):
Why can't we just use CONFIG_TFM_BOARD?

CMakeLists.txt
erwango: @galak, cf relative discussion here: https://github.com/zephyrproject-rtos/zephyr/pull/30062#discussion_r524413356

CMakeLists.txt
```diff
@@ -34,6 +34,25 @@ function(trusted_firmware_build)
...
+    set(TFM_PLATFORM_DIR "mps2/an521")
+  elseif(CONFIG_TFM_BOARD_MUSCA_B1)
+    set(TFM_PLATFORM_DIR "musca_b1")
+  elseif(CONFIG_TFM_BOARD_NRF5340DK_NRF5340)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf5340dk_nrf5340_cpuapp")
+  elseif(CONFIG_TFM_BOARD_NRF5340PDK_NRF5340)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf5340pdk_nrf5340_cpuapp")
+  elseif(CONFIG_TFM_BOARD_NRF9160DK_NRF9160)
+    set(TFM_PLATFORM_DIR "nordic_nrf/nrf9160dk_nrf9160")
+  elseif(CONFIG_TFM_BOARD_BOARD_NUCLEO_L552ZE_Q)
```
erwango, line 50:
```suggestion
  elseif(CONFIG_TFM_BOARD_NUCLEO_L552ZE_Q)
```

erwango (changes requested):

ioannisg:
@galak the summary is as follows:

- @tejlmand  believes this is the right approach, because effectively the TFM_PLATFORM_DIR are build paths for TF-M and the information belongs to CMake (not Kconfig)
- the problem is that CMakeLists.txt is _still_ present in the module side, not in Zephyr side but, according to my understanding (CC @tejlmand) this is temporary. This will move to Zephyr, so we will follow the convention of moving shim layers inside the Zephyr tree.

So this PR could be treated as an intermediate step and it should be accepted IMO. Otherwise we can block the PR until the framework for moving CMake code from module to Zephyr is available, which I don't want to do (but can also accept if you insist).

galak:
> @galak the summary is as follows:
> 
> * @tejlmand  believes this is the right approach, because effectively the TFM_PLATFORM_DIR are build paths for TF-M and the information belongs to CMake (not Kconfig)
> * the problem is that CMakeLists.txt is _still_ present in the module side, not in Zephyr side but, according to my understanding (CC @tejlmand) this is temporary. This will move to Zephyr, so we will follow the convention of moving shim layers inside the Zephyr tree.
> 
> So this PR could be treated as an intermediate step and it should be accepted IMO. Otherwise we can block the PR until the framework for moving CMake code from module to Zephyr is available, which I don't want to do (but can also accept if you insist).

If the intent is to phase out CONFIG_TFM_BOARD, that is fine, but I don't see any reason to introduce new TFM specific Kconfig options, there should be enough already with CONFIG_BOARD_* or some CONFIG_SOC_* that can be used here.

ioannisg:
Closing this PR, as it is not applicable any more, as we fully-reworked the CMakeLists for tFM

PR #31, opened 2020-11-20
CMakeList.txt: disable TEST_NS

agansari:
Removes TEST_NS, eqivalent to -DTEST_NS=OFF set as default. This
setting affects tfm_ns.bin image which is not used in Zephyr.
TEST_NS adds regression tests to TFM's non-secure image, does not
affect independent non-secure applications.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ function(trusted_firmware_build)
   endif()
 
   if(DEFINED TFM_REGRESSION)
-    set(TFM_REGRESSION_ARG -DTEST_S=ON -DTEST_NS=ON)
+    set(TFM_REGRESSION_ARG -DTEST_S=ON)
   endif()
 
   if(DEFINED TFM_CMAKE_BUILD_TYPE)


galak (approved):

agansari:
@oyvindronningstad maybe add the removed define only when https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/22 is built.

oyvindronningstad:
> 
> 
> @oyvindronningstad maybe add the removed define only when #22 is built.

Yep, that sounds like a plan, something like:

    if(DEFINED TFM_REGRESSION)
      set(TFM_REGRESSION_ARG -DTEST_S=ON)
      if(TFM_NATIVE_NS)
        list(APPEND TFM_REGRESSION_ARG -DTEST_NS=ON)
      endif()
    endif()

oyvindronningstad (approved):

ioannisg (commented):
Fine with me, as well, but I wonder if it would make sense to have separate Kconfig options that will enable TEST_S and TEST_NS. Currently we have config TFM_REGRESSION which enables both.

Would we ever need to have TFM_NS on? Might do so if we want to run TFM_REGRESSION tests via Zephyr build system, no?

@microbuilder ?

oyvindronningstad:
> Fine with me, as well, but I wonder if it would make sense to have separate Kconfig options that will enable TEST_S and TEST_NS. Currently we have config TFM_REGRESSION which enables both.
> 
> Would we ever need to have TFM_NS on? Might do so if we want to run TFM_REGRESSION tests via Zephyr build system, no?

FYI that's what the comments above regarding #22 are about. Until #22 there is no reason to enable TEST_NS. At that point, it's relevant if NATIVE_NS is enabled, like you say.

ioannisg (approved):

PR #33, opened 2020-11-24
trusted-firmware-m: nrf5340: Fix UART pins

oyvindronningstad:
Fix UART1 (secure) pins for nrf5340dk_nrf5340_cpuapp which are different
from the pdk's pins.

Upstream PR: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/7177

### DIFF
--- a/trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340dk_nrf5340_cpuapp/RTE_Device.h
+++ b/trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340dk_nrf5340_cpuapp/RTE_Device.h
@@ -40,13 +40,13 @@
 #define   RTE_USART1                    1
 //   <h> Pin Selection (0xFFFFFFFF means Disconnected)
 //     <o> TXD
-#define   RTE_USART1_TXD_PIN            25
+#define   RTE_USART1_TXD_PIN            33
 //     <o> RXD
-#define   RTE_USART1_RXD_PIN            26
+#define   RTE_USART1_RXD_PIN            32
 //     <o> RTS
-#define   RTE_USART1_RTS_PIN            0xFFFFFFFF
+#define   RTE_USART1_RTS_PIN            11
 //     <o> CTS
-#define   RTE_USART1_CTS_PIN            0xFFFFFFFF
+#define   RTE_USART1_CTS_PIN            10
 //   </h> Pin Configuration
 // </e> USART (Universal synchronous - asynchronous receiver transmitter) [Driver_USART1]
 


oyvindronningstad:
cc @anangl @magnev

trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340dk_nrf5340_cpuapp/RTE_Device.h
```diff
@@ -40,13 +40,13 @@
...
+#define   RTE_USART1_TXD_PIN            33
 //     <o> RXD
-#define   RTE_USART1_RXD_PIN            26
+#define   RTE_USART1_RXD_PIN            32
 //     <o> RTS
-#define   RTE_USART1_RTS_PIN            0xFFFFFFFF
+#define   RTE_USART1_RTS_PIN            11
 //     <o> CTS
-#define   RTE_USART1_CTS_PIN            0xFFFFFFFF
+#define   RTE_USART1_CTS_PIN            10
```
anangl, line 49: These pins are [assigned to UART in the nrf5340dk_nrf5340_cpunet board](https://github.com/zephyrproject-rtos/zephyr/blob/1d4c36486c94212478bf222e12bf463affb77d55/boards/arm/nrf5340dk_nrf5340/nrf5340dk_nrf5340_cpunet.dts#L93-L96), so I'm not convinced we should use them for the secure UART in TF-M.

trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340dk_nrf5340_cpuapp/RTE_Device.h
```diff
@@ -40,13 +40,13 @@
...
 //   <h> Pin Selection (0xFFFFFFFF means Disconnected)
 //     <o> TXD
-#define   RTE_USART1_TXD_PIN            25
+#define   RTE_USART1_TXD_PIN            33
 //     <o> RXD
-#define   RTE_USART1_RXD_PIN            26
+#define   RTE_USART1_RXD_PIN            32
 //     <o> RTS
-#define   RTE_USART1_RTS_PIN            0xFFFFFFFF
+#define   RTE_USART1_RTS_PIN            11
```
anangl, line 47: The secure UART in TF-M does not use flow control, so we don't need to assign any pins as RTS and CTS lines. Actually, even RXD is not needed, as the console is output-only.

trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340dk_nrf5340_cpuapp/RTE_Device.h
anangl: I see that for the PDK we already got such conflict: pins P0.25 and P0.26 are used by the secure UART https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/74f148fee35d7019f5b74ebb562f78cf6f988804/trusted-firmware-m/platform/ext/target/nordic_nrf/nrf5340pdk_nrf5340_cpuapp/RTE_Device.h#L42-L45
and by the [UART in the nrf5340pdk_nrf5340_cpunet board](https://github.com/zephyrproject-rtos/zephyr/blob/1d4c36486c94212478bf222e12bf463affb77d55/boards/arm/nrf5340dk_nrf5340/nrf5340pdk_nrf5340_cpunet.dts#L93-L94) in Zephyr. I guess that should be fixed.

oyvindronningstad: What happens if both cores use the same UART pins?

anangl: > What happens if both cores use the same UART pins?

Only the core to which those pins are assigned (via `PIN_CNF[n].MCU_SEL`) will be able to access them.

PR #35, opened 2020-12-02
TF-M upmerge to upstream release v1.2.0

ioannisg:
Synchronizing TF-M module with upstream project on release v1.2.0.


agansari (approved):
Tested with https://github.com/zephyrproject-rtos/zephyr/pull/30013 works as expected.

ioannisg:
Smoke test; it works as-is for AN521, no need to patch the CMakeLists.txt.
Need to carry back the patches for the Nucleo platform. Will do that and then ask @erwango  to take a look.

galak (approved):

oyvindronningstad:
Should a6e0516f8e04f6ac65b01f4fb3255122a1eb4453 be reverted?

ioannisg:
> Should [a6e0516](https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/a6e0516f8e04f6ac65b01f4fb3255122a1eb4453) be reverted?

It is "reverted" because we do the "shapshot" approach for the synchronization. So if it is merged upstream it is already part of the synchronization commit.

oyvindronningstad (approved):

microbuilder (approved):

ioannisg:
> Smoke test; it works as-is for AN521, no need to patch the CMakeLists.txt.
> Need to carry back the patches for the Nucleo platform. Will do that and then ask @erwango to take a look.

This is now done, and I've verified the integration sample (tfm-ipc) on nucleo_l552ze_q_ns.

erwango (approved):

PR #40, opened 2021-01-15
CMakeLists.txt: enable/disable partitions

hakonfam:
Parse kconfig symbols to enable/disable partitions
and pass the arguments to the tfm build function.

Signed-off-by: Håkon Øye Amundsen <haakon.amundsen@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+# List of all partitions supported by TF-M
+# Name must match name in 'trusted-firmware-m/config/config_default.cmake'
+set(TFM_VALID_PARTITIONS
+  TFM_PARTITION_PROTECTED_STORAGE
+  TFM_PARTITION_INTERNAL_TRUSTED_STORAGE
+  TFM_PARTITION_CRYPTO
+  TFM_PARTITION_INITIAL_ATTESTATION
+  TFM_PARTITION_PLATFORM
+  TFM_PARTITION_AUDIT_LOG
+  )
+
 # Adds trusted-firmware-m as an external project.
 # Also creates a target called 'tfm_api'
 # which can be linked into the app.
@@ -18,6 +29,7 @@
 # ISOLATION_LEVEL: The TF-M isolation level to use
 # REGRESSION: Boolean if TF-M build includes building the TF-M regression tests
 # BL2: Boolean if the TF-M build uses MCUboot. Default: True
+# ENABLED_PARTITIONS: List of TFM partitions to enable.
 #
 # Example usage:
 #
@@ -28,15 +40,24 @@
 #                        ISOLATION_LEVEL 2
 #                        REGRESSION
 #                        BL2 True
-#                        BUILD_PROFILE profile_small)
+#                        BUILD_PROFILE profile_small
+#                        ENABLED_PARTITIONS TFM_PARTITION_PLATFORM TFM_PARTITION_CRYPTO)
 function(trusted_firmware_build)
   set(options IPC REGRESSION)
   set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
-  cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "" ${ARGN})
+  set(multiValueArgs ENABLED_PARTITIONS)
+  cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
-  if(NOT DEFINED TFM_BL2)
-    set(TFM_BL2 True)
-  endif()
+  foreach(partition ${TFM_VALID_PARTITIONS})
+    list(FIND TFM_ENABLED_PARTITIONS ${partition} idx)
+    if (idx EQUAL -1)
+      set(val "OFF")
+    else()
+      set(val "ON")
+    endif()
+    list(APPEND TFM_PARTITIONS_ARGS -D${partition}=${val})
+  endforeach()
+
   set(TFM_BL2_ARG "-DBL2=${TFM_BL2}")
 
   if(DEFINED TFM_IPC)
@@ -127,6 +148,7 @@ function(trusted_firmware_build)
                ${TFM_PROFILE_ARG}
                -DTFM_TEST_REPO_PATH=${ZEPHYR_TFM_MODULE_DIR}/tf-m-tests
                -DMCUBOOT_PATH=${ZEPHYR_TFM_MODULE_DIR}/../tfm-mcuboot
+               ${TFM_PARTITIONS_ARGS}
     BUILD_ALWAYS True
     USES_TERMINAL_BUILD True
     BUILD_BYPRODUCTS ${BUILD_BYPRODUCTS}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -28,14 +40,30 @@
...
-  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
+  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE
+    BUILD_PROFILE ${tfm_partitions})
   cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "" ${ARGN})
 
-  if(NOT DEFINED TFM_BL2)
-    set(TFM_BL2 True)
+  # Set all arguments which enable/disable the secure partitions.
+  foreach(partition ${tfm_partitions})
+    set(val ${TFM_${partition}})
```
tejlmand, line 58: This means that using this function, it is now a requirement to specify ALL partitions.
Is that really needed ?
Why not default each partition to `OFF`, and then use the argument name: `PARTITIONS` in a multi args argument, see other comment.

CMakeLists.txt
```diff
@@ -28,14 +40,30 @@
...
+#                        TFM_PARTITION_INTERNAL_TRUSTE_STORAGE ON
+#                        TFM_PARTITION_CRYPTO ON
+#                        TFM_PARTITION_INITIAL_ATTESTATION ON
+#                        TFM_PARTITION_PLATFORM ON
+#                        TFM_PARTITION_AUDIT_LOG ON)
 function(trusted_firmware_build)
   set(options IPC REGRESSION)
-  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
+  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE
+    BUILD_PROFILE ${tfm_partitions})
```
tejlmand, line 53: I suggest to use multi value argument here, like:
```
  set(oneValueArgs BINARY_DIR BOARD BL2 ISOLATION_LEVEL CMAKE_BUILD_TYPE BUILD_PROFILE)
  set(multiValueArgs PARTITIONS)
  cmake_parse_arguments(TFM "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
```

then caller can call this function as:
```
 trusted_firmware_build(.... PARTITIONS TFM_PARTITION_PROTECTED_STORAGE TFM_PARTITION_INTERNAL_TRUSTE_STORAGE TFM_PARTITION_INITIAL_ATTESTATION ...)
```

for the partitions to enable.

tejlmand (changes requested):
Unless there are very specific reasons to request explicit decision making for each partition, then I think partitions should default be off (in the CMake code, Kconfig may decide differently), and user or the sample can then choose partitions to enable.

This also allows future partitions to be added by TFM, without having to update code, as those simply will default to `OFF`.

CMakeLists.txt
```diff
@@ -3,6 +3,17 @@
...
+# List of all partitions supported by TF-M
+# Name must match name in 'trusted-firmware-m/config/config_default.cmake'
+set(tfm_partitions
+  TFM_PARTITION_PROTECTED_STORAGE
+  TFM_PARTITION_INTERNAL_TRUSTED_STORAGE
+  TFM_PARTITION_CRYPTO
+  TFM_PARTITION_INITIAL_ATTESTATION
+  TFM_PARTITION_PLATFORM
+  TFM_PARTITION_AUDIT_LOG
+  )
```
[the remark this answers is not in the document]
hakonfam: Isn't that a bit overkill? If we do as you say elsewhere, and we set the default to OFF, then the only thing we have to do if a new partition is added is to add the entry in the list during the upmerge. I thought that we wanted to avoid unnecessary `execute_process` to make the execution time of the configure stage faster.

Some synchronization would still have to be done, since the name of the existing partitions might change.

CMakeLists.txt
hakonfam: That sounds much better.

CMakeLists.txt
hakonfam: We have to explicitly state `OFF` for all the partitions, so we need some more logic in the function in this case, but I agree that it is better seen from the callers perspective.

CMakeLists.txt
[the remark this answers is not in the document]
tejlmand: > Some synchronization would still have to be done, since the name of the existing partitions might change.

Right, especially on the Kconfig side, as you are using the same symbols there.
I'm fine with keeping a list, only proposed automation, to minimize the burden on keeping the names in sync for the module maintainers.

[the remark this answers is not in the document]
tejlmand: No need to change.

CMakeLists.txt
tejlmand: you already have a `foreach` so would be very simple to adjust to something like:
```
set(partitions_off ${TFM_VALID_PARTITIONS})  # Defaulting everything to off.
list(REMOVE_ITEM partitions_off ${TFM_PARTITIONS}) # Those are provided by the user in the call, so remove from the off list

foreach(partition ${partitions_off})
    list(APPEND TFM_PARTITIONS_ARGS -D${partition}=OFF)
endforeach()

foreach(partition ${TFM_PARTITIONS})
    list(APPEND TFM_PARTITIONS_ARGS -D${partition}=ON)
endforeach()
```

tejlmand: and then you can of course remove:
```
 foreach(partition ${TFM_VALID_PARTITIONS})
    list(FIND TFM_ENABLED_PARTITIONS ${partition} idx)
    if (idx EQUAL -1)
      set(val "OFF")
    else()
      set(val "ON")
    endif()
    list(APPEND TFM_PARTITIONS_ARGS -D${partition}=${val})
  endforeach()
```

tejlmand: never mind.
I see now that I was already looking at the updated code.

CMakeLists.txt
```diff
@@ -201,6 +223,14 @@ if (CONFIG_BUILD_WITH_TFM)
     set(TFM_PROFILE_ARG BUILD_PROFILE ${CONFIG_TFM_PROFILE})
   endif()
 
+  # Enable TFM partitions as specified in Kconfig
+  set(TFM_ENABLED_PARTITIONS_ARG ENABLED_PARTITIONS)
```
tejlmand, line 227: I would remove this line, and place the argument name `ENABLED_PARTITIONS` in the call for better readability.

It is ok for the following list of partitions to be empty (and `list(APPEND)` will create the list if not yet present ).
```suggestion
```

CMakeLists.txt
```diff
@@ -209,6 +239,7 @@ if (CONFIG_BUILD_WITH_TFM)
     ${TFM_BL2_ARG}
     ${TFM_IPC_ARG}
     ${TFM_REGRESSION_ARG}
+    ${TFM_ENABLED_PARTITIONS_ARG}
```
tejlmand, line 242: As mentioned previously, I believe this provides better readability in terms of `argument name` and `value list.
```suggestion
    ENABLED_PARTITIONS ${TFM_ENABLED_PARTITIONS_ARG}
```

tejlmand (approved):
Looking good.
Consider to take in the suggestion on the argument list.

CMakeLists.txt
hakonfam: True, this is better, thanks!

hakonfam:
ping @ioannisg @oyvindronningstad

oyvindronningstad (approved):

microbuilder (approved):

ioannisg (approved):

ioannisg:
@hakonfam pls, close and re-direct to the main tree.

hakonfam:
Included in https://github.com/zephyrproject-rtos/zephyr/pull/31349

PR #44, opened 2021-02-03
CMakeLists.txt: Pass CMAKE_BUILD_TYPE from Zephyr build

oyvindronningstad:
to the TFM build.


microbuilder (approved):

hakonfam (approved):

CMakeLists.txt
```diff
@@ -198,6 +198,9 @@ if (CONFIG_BUILD_WITH_TFM)
   if (CONFIG_TFM_PROFILE)
     set(TFM_PROFILE_ARG BUILD_PROFILE ${CONFIG_TFM_PROFILE})
   endif()
+  if (DEFINED CMAKE_BUILD_TYPE)
```
tejlmand, line 201: Zephyr don't use the `CMAKE_BUILD_TYPE` and it's not advised to use it, as Zephyr uses Kconfig for controlling what is normally controlled through the app.

Although it may be used by the app for other purposes, then I would really like to know why this is needed, as the way an app developer could use the CMAKE_BUILD_TYPE is not the same as how TFM might use it.

See more here:
https://github.com/zephyrproject-rtos/zephyr/issues/18652
https://github.com/zephyrproject-rtos/zephyr/pull/18777

As note, if user specifies CMAKE_BUILD_TYPRE and there is a mismatch between the Kconfig optimization setting, and the CMAKE_BUILD_TYPE, the user will be presented with the warning:
https://github.com/zephyrproject-rtos/zephyr/blob/b6a4db3766936d738ed84f9f6a4ee19f586ce062/CMakeLists.txt#L1374-L1375
```
      The CMake build type was set to '${CMAKE_BUILD_TYPE}', but the optimization flag was set to '${OPTIMIZATION_FLAG}'.
      This may be intentional and the warning can be turned off by setting the CMake variable 'NO_BUILD_TYPE_WARNING'"
```
Thus we should not encourage users to start using the CMAKE_BUILD_TYPE.

I propose that if there is a need to control the CMAKE_BUILD_TYPE for TF-M then do one of the following:
- Create a dedicated Kconfig setting for TF-M build type, which allow controlling the build type independently of Zephyr
   The TF-M Build type could have default value based on the Zephyr optimization setting
- Deduce the TF-M Build type automatically from the Zephyr optimization setting.

ioannisg (approved):

CMakeLists.txt
oyvindronningstad: Thanks a lot for the background info. @tejlmand!

I'll create a dedicated Kconfig. We can set it based on the Zephyr Kconfig configuration at some point.

oyvindronningstad: done @tejlmand

oyvindronningstad:
CMakeLists.txt has been moved into zephyr.

PR #46, opened 2021-02-11
CMakeLists.txt: Use new TFM_BL2 kconfig

oyvindronningstad:
Change the symbol usage.


hakonfam (approved):

microbuilder (approved):

CMakeLists.txt
```diff
@@ -187,9 +187,9 @@ if (CONFIG_BUILD_WITH_TFM)
   if (CONFIG_TFM_REGRESSION)
     set(TFM_REGRESSION_ARG REGRESSION)
   endif()
-  if (CONFIG_TFM_BL2_TRUE)
+  if (CONFIG_TFM_BL2)
```
tejlmand, line 190: Now that `CONFIG_TFM_BL2` is either true or false, then it is possible to remove this `if() ... else()` and just do following below:

```
 trusted_firmware_build(
    BINARY_DIR ${CMAKE_BINARY_DIR}/tfm
    BOARD ${CONFIG_TFM_BOARD}
    BL2 ${CONFIG_TFM_BL2}
    ...
```

and in line 38:
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/f3608a012192ad2ef48ef7a8df11666f176662c5/CMakeLists.txt#L38
change to:
```
    set(TFM_BL2 False)
```
cause an enabled Kconfig setting will also be true, and disable either be empty (when not explicitly set) or false.
(which means you only need to handle the `empty` == not defined case == false)

This cleans up the code.

You may also choose the change `TFM_BL2` to an option flag instead.

tejlmand (changes requested):
Think we should reduce code, now that we have the possibility.

CMakeLists.txt
oyvindronningstad: I looked at it, but I want to avoid the ambiguity of empty vs false. I did manage to make it smaller, see the zephyr PR.

oyvindronningstad:
CMakeLists.txt has been moved into zephyr.

CMakeLists.txt
tejlmand: LGTM :+1:

PR #47, opened 2021-03-12
Update psa-arch-tests to support running PSA tests

oyvindronningstad:
Make psa-arch-tests support CROSS_COMPILE like TFM does. This is needed to build PSA tests with the Zephyr SDK.

### DIFF
--- a/psa-arch-tests/api-tests/CMakeLists.txt
+++ b/psa-arch-tests/api-tests/CMakeLists.txt
@@ -109,6 +109,11 @@ list(APPEND PSA_TOOLCHAIN_SUPPORT
         HOST_GCC
 )
 
+# list of supported CROSS_COMPILE toolchains
+list(APPEND CROSS_COMPILE_TOOLCHAIN_SUPPORT
+        GNUARM
+)
+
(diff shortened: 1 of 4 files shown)


microbuilder:
@oyvindronningstad  Even if it's basically just an update to `psa-arch-tests`, there are a lot of files here to review. Can you update the PR description with some details on what this adds/does, and how to test?

oyvindronningstad:
@microbuilder sure I can do that. I'll hold off now until the upmerge goes in since they are in conflict.

oyvindronningstad:
@microbuilder This is ready to review now. I opted not to upmerge to the psa-arch-tests master here, so this is now just the one change that is needed for psa tests (the CROSS_COMPILE support).

microbuilder (approved):

tejlmand (approved):

oyvindronningstad:
@ioannisg Can you take a look?

ioannisg (approved):
looks good to me, thanks
I think we should prefer if we can get the corresponding upstream patch merged, right?
Then we could just synchronize. Correct @oyvindronningstad ?

PR #48, opened 2021-03-22
TFM: Update to v1.3.0-rc1

oyvindronningstad:


galak (approved):
Probably good to sync to TF-Mv1.3.0-RC1 now that its out.

oyvindronningstad:
> 
> 
> Probably good to sync to TF-Mv1.3.0-RC1 now that its out.

Done

galak (commented):
Some how we seem to be missing:
```
 config/build_type/debug.cmake                                     |   10 ----------
 config/build_type/minsizerel.cmake                                |   10 ----------
 config/build_type/release.cmake                                   |   11 -----------
```

oyvindronningstad:
> 
> 
> Some how we seem to be missing:
> 
> ```
>  config/build_type/debug.cmake                                     |   10 ----------
>  config/build_type/minsizerel.cmake                                |   10 ----------
>  config/build_type/release.cmake                                   |   11 -----------
> ```

@galak Thanks, they were ignored by tfm's .gitignore.

microbuilder (approved):

ioannisg (changes requested):
I think this needs a README update. Let me know if i missed that

ioannisg:
I have a feeling that we do not need, any more, the fixes for the STM scripts, since, at least, the TFM_UPdate.sh is not used any more, by default. It is not a problem to keep the patches, though, since the scripts still exist.  Just a not for future cleanup.

ioannisg (approved):

galak:
> I have a feeling that we do not need, any more, the fixes for the STM scripts, since, at least, the TFM_UPdate.sh is not used any more, by default. It is not a problem to keep the patches, though, since the scripts still exist. Just a not for future cleanup.

I wondered about upstreaming some of those changes.  But good to know that they might not be needed.

oyvindronningstad:
> 
> 
> > I have a feeling that we do not need, any more, the fixes for the STM scripts, since, at least, the TFM_UPdate.sh is not used any more, by default. It is not a problem to keep the patches, though, since the scripts still exist. Just a not for future cleanup.
> 
> I wondered about upstreaming some of those changes. But good to know that they might not be needed.

They actually were needed. I tried first without, but the CI failed.

PR #49, opened 2021-04-11
Tfm upmerge to v.1.3.0

ioannisg:
Upmerge to v1.3.0, recently released.
psa-arch-tests is updated to the SHA that is the default in TF-M 1.3.0
MCUboot for TFM stayed at v1.7.2


galak (approved):

microbuilder (approved):
Thanks for updating this.

oyvindronningstad (approved):

PR #51, opened 2021-05-19
platform: Enable instruction cache on platforms with nRF SoCs

anangl:
Instruction cache in both nRF5340 and nRF9160 is disabled by default
and can be enabled only from secure code. Extend configuration routines
in targets that use these SoCs with statements that enable the cache.
The affected targets are:
- lairdconnectivity/bl5340_dvk_cpuapp
- nordic_nrf/nrf5340dk_nrf5340_cpuapp
- nordic_nrf/nrf9160dk_nrf9160

Upstream PR: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10050


oyvindronningstad (approved):

galak (approved):

ioannisg (approved):

galak:
We should probably wait on merging this if the plan is to re-sync with upstream TFM to get the mbedtls 2.26.0 update.

galak:
Marked DNM for now so we merge this post tfm update for mbedtls

ioannisg:
> We should probably wait on merging this if the plan is to re-sync with upstream TFM to get the mbedtls 2.26.0 update.

@galak I was about to say the same thing, thanks!

anangl:
Superseded by #52.

PR #53, opened 2021-05-24
[DNM] Tfm upmerge to v.1.3.99 with fixes for v2.6

galak:


trusted-firmware-m/platform/ext/target/nxp/lpcxpresso55s69/target_cfg.c
```diff
@@ -435,8 +435,8 @@ int32_t ppc_init_cfg(void)
 
     AHB_SECURE_CTRL->SEC_CTRL_APB_BRIDGE[0].SEC_CTRL_APB_BRIDGE1_MEM_CTRL2 =
         (0x33003333U) |                                                                     /* Bits have to be set to '1' according to UM.*/
-        AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_FLASH_CTRL_RULE(0x3U) |              /* Flash controller = S*/
-        AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_PRINCE_RULE(0x3U);                   /* PRINCE = S */
+        AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_FLASH_CTRL_RULE(0x0U) |              /* Flash controller = S*/
+        AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_PRINCE_RULE(0x0U);                   /* PRINCE = S */
```
microbuilder, line 439: PRINCE should still be set to S (0x3) ... only Flash Controller needs to change, and the comment should be updated to say `NS`

```
AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_FLASH_CTRL_RULE(0x0U) |              /* Flash controller = NS*/
AHB_SECURE_CTRL_SEC_CTRL_APB_BRIDGE1_MEM_CTRL2_PRINCE_RULE(0x3U);                   /* PRINCE = S */
```

microbuilder (changes requested):

trusted-firmware-m/platform/ext/target/nxp/lpcxpresso55s69/target_cfg.c
microbuilder: This flag limits access to the ROM-based flash controller, which has correctly been set to S in my opinion as a security precaution to prevent access to the contents of flash memory on the secure side.

https://github.com/zephyrproject-rtos/zephyr/pull/35599 fixes this issue without lowering the flash restrictions, by disabling flash access on the NS side, which is the correct course of action in my opinion.

PR #55, opened 2021-06-29
trusted-firmware-m: partition: Add an example partition

urutva:
Add an example partition which provides a secure SHA-256 hash service
that computes the SHA-256 hash of a given message and returns the hash.
In addition, it also contains a templated example which can be used as
a starting point for developing secure services. The templated example
service shows how to start a platform timer and implement a timer IRQ
handler.

Signed-off-by: Devaraj Ranganna <devaraj.ranganna@linaro.org>


SebastianBoe (changes requested):
TFM have since changed how custom secure partitions should be added.

This PR needs to be updated to use this new mechanism.

https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10562

urutva:
@SebastianBoe @microbuilder I'm closing this PR in favour of PR by @oyvindronningstad https://github.com/zephyrproject-rtos/zephyr/pull/36595 which needs no changes in TF-M to add a new secure partition. I've created a templated example that helps in adding new secure services. I can add those changes to https://github.com/zephyrproject-rtos/zephyr/pull/36595 or create a new PR once it is merged.

PR #57, opened 2021-07-14
Tfm upmerge to v.1.3.99 post 2.6

ioannisg:
Regular upmerge for TF-M module

- Still in v1.3.99, however, with some good fixes
- Several downstream patches are removed as obsolete (corresponding patches are merged upstream)


microbuilder (approved):
Tested on `lpcxpresso55s69_ns` and `mps2_an521_ns`

galak (approved):

PR #58, opened 2021-08-06
trusted-firmware-m: platform: lpcxpresso55s69: Update SDK

galak:
Update LPCXPRESSO55S69 platform SDK to version 2.10.0 to be
in sync with Zephyr usage of SDK 2.10.0.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


microbuilder (approved):
Tested with:

```bash
$ west build -p auto -b lpcxpresso55s69_ns zephyr/samples/hello_world/
```

and:

```bash
$ west build -p auto -b lpcxpresso55s69_ns zephyr/samples/tfm_integration/psa_crypto/
```

flashing via:

```bash
$ JLinkExe -device lpc55s69 -if swd -speed 2000 -autoconnect 1
r
erase
loadfile build/tfm_merged.hex
r
go
```

dleach02 (approved):

ioannisg:
@galak @microbuilder @dleach02 ideally we should be getting the NXP HAL for building TFM directly from Zephyr's module, instead of checking in the files here. This was supposed to be a short-term solution. I think we should fix this before the LTS release.

PR #60, opened 2021-08-16
Tfm upmerge to v.1.4.0

ioannisg:
TF-M module synchronization with upstream TF-M v1.4.0 release.


microbuilder (commented):
Test with `mps2_an521_ns` and works, but `lpcxpresso55s69_ns` predictably fails since it isn't finding the NXP SDK files that are downloaded at compile time with upstream TF-M.

This PR seems good, but I'll need to update https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/60 to include the missing NXP files for TF-M to work on that board.

Verified good wiith the following, though:

```bash
$ west build -p auto -b mps2_an521_ns zephyr/samples/tfm_integration/psa_crypto/ -t run
```

EDIT: Tested with `lpcxpesso55s69_ns` as well after https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/60/commits/adc2866c9f2a6efd4a1eb4b75b79ae9d5c30ccb9

microbuilder:
Above test requires https://github.com/zephyrproject-rtos/zephyr/pull/38149

psa-arch-tests/api-tests/CMakeLists.txt
```diff
@@ -582,14 +582,18 @@ add_custom_target(
...
+
 # Check for supported toolchain/s
 if(${TOOLCHAIN} IN_LIST PSA_TOOLCHAIN_SUPPORT)
 	if (DEFINED CROSS_COMPILE)
 		if(NOT (${TOOLCHAIN} IN_LIST CROSS_COMPILE_TOOLCHAIN_SUPPORT))
 			message(FATAL_ERROR "[PSA] : Error: CROSS_COMPILE not supported for this toolchain, supported toolchain are : ${CROSS_COMPILE_TOOLCHAIN_SUPPORT}")
 		endif()
 	endif()
-	include(${PSA_ROOT_DIR}/tools/cmake/compiler/${TOOLCHAIN}.cmake)
+    include(${PSA_TOOLCHAIN_FILE})
```
tejlmand, line 596: why the extra indentation here ?
It doesn't align with the `endif()` above.

SebastianBoe: Looks like I need to configure my editor, it was showing the same indentation.

Will fix.

I see that this project uses a mix of tabs and spaces for indentation, whereas Zephyr uses only whitespace.

Not sure how to configure emacs for a mix of tabs and spaces ...

SebastianBoe: cropped

SebastianBoe: I guess I could edit in whitespace-mode and ensure I use the same blend of tabs and whitespaces as the line above.

SebastianBoe: ... or we could accept that upstream does not have sane management of whitespace and accept the natural indentation issues that follow.

tejlmand: > I see that this project uses a mix of tabs and spaces for indentation, whereas Zephyr uses only whitespace.

😞 not much we can do then, we shouldn't cleanup their mess.
Of course we should probably just keep the existing indentation when modifying an already existing line, but then it still becomes tricky if we add new lines.

Will leave the choice up to you, as regardless of what we do, it will be a mess.

Thanks for looking into this.

tejlmand (approved):
Approved. Thanks for looking into their indentation mess.

SebastianBoe (approved):

SebastianBoe:
We should upmerge to the latest release, which is TF-M 1.4.1.

SebastianBoe:
Deprecated by https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/61

PR #61, opened 2021-09-29
V.1.4.1 upmerge

SebastianBoe:
TF-M module synchronization with upstream TF-M v1.4.1 release.


d3zd3z (approved):
Assuming this is changes coming from upstream, this looks good.

SebastianBoe:
There are a couple of commits that are not from upstream:

https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/61/commits/4c49d272b76245b8a403b23930adfbfa9d71eb01
https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/61/commits/adc2866c9f2a6efd4a1eb4b75b79ae9d5c30ccb9

The SDK update I don't know the story of.

But the toolchain change is simply a necessary patch to be able to use TF-M. Will try to patch upstream so it is not needed at a later time.

SebastianBoe:
The SDK update is a natural consequence of us copying the SDK into the TF-M repo instead of downloading it as upstream TF-M does.

carlescufi (approved):

PR #63, opened 2021-12-01
TF-M 1.5

SebastianBoe:
Update from 1.4.1 to 1.5.0.

We are following TF-M quite closely.

We have one trivial doc fix where we add the 1.4.1 release to docs/releases/index.rst.
This will be upstreamed soon.

We have the west module.yml file.

And then the controversial NXP SDK. I think it is added because we need to avoid downloading the SDK on every build, but I need to double-check that this is still true and that we are not accidentally downloading the SDK anyway.


SebastianBoe:
DNM.

Needs testing and also the NXP SDK patch needs to be checked. Maybe it is not needed.

joerchan:
Could you update the commit message to the tag being merged, instead of the release branch, which can be moved.
TF-Mv1.5.0 is the tag.

SebastianBoe:
Done

joerchan (approved):

microbuilder (changes requested):
The NXP SDK situation is something we weren't able to change upstream, where a change was made to remove the commited SDK files and download every build.  Adding the files here as part of TF-M is the wrong approach, though (https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/63/commits/2cc2948a0e698ca96a730ab2a341af0f734d5992). We need to set `TFM_PLATFORM_NXP_HAL_FILE_PATH` to point to https://github.com/zephyrproject-rtos/hal_nxp (which is at 2.10.0 as of this comment), overriding the default download action.

For details see the discission here: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10225

The specific logic is defined upstream as follows:

https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/tree/platform/ext/target/nxp/lpcxpresso55s69/CMakeLists.txt#n12

```cmake
# Pull drivers from NXP git
if (TFM_PLATFORM_NXP_HAL_FILE_PATH STREQUAL "DOWNLOAD")
    set(NXP_HAL_FILE_PATH ${PLATFORM_DIR}/ext/target/nxp)
    Message("-- Pulling MCUxpresso NXP SDK drivers from https://github.com/NXPmicro/mcux-sdk")
    include(pull_drivers.cmake)
# Use specified path to NXP hal folder provided by user
else()
    set(NXP_HAL_FILE_PATH ${TFM_PLATFORM_NXP_HAL_FILE_PATH})
endif()
```

We set this for Zephyr with 1.4.0 via:

```
zephyr/modules/trusted-firmware-m/CMakeLists.txt
208:    set(TFM_PLATFORM_NXP_HAL_FILE_PATH ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/platform/ext/target/nxp/)
209:    list(APPEND TFM_CMAKE_ARGS -DTFM_PLATFORM_NXP_HAL_FILE_PATH=${TFM_PLATFORM_NXP_HAL_FILE_PATH})
```

joerchan:
> The NXP SDK situation is something we weren't able to change upstream, where a change was made to remove the commited SDK files and download every build.

@microbuilder Are you saying that the SDK was removed in TF-M without your approval? Is this something that will happen to other HALs?

Using `TFM_PLATFORM_NXP_HAL_FILE_PATH` looks like the right approach I agree. I'm curious why nobody from NXP has corrected this though, they should be invested in their own platform support.
From what I can tell there is already an attempt to use this handling, https://github.com/zephyrproject-rtos/zephyr/commit/d42671737e6be281b69d3e67d9723f4340c7e8c7.
Why we still ended up with compilation errors doesn't make sense to me then. But maybe someone from NXP can clear that up after the upmerge, and we continue with the upmerge as it is, fixing the NXP SDK doesn't have to be part of the upmerge IMO.

microbuilder:
@joerchan Initially, I added the LPC55S59 support to TF-M since we needed a board that was easier to purchase than the Musca boards, etc. The required NXP SDK files were part of TF-M at that time, which was how other devices handled this. NXP picked up maintenance for it at a later data (originally they were working purely on a fork they maintained with their own IDE-based build system), and part of the changes they introduced was removing the SDK files from TF-M and downloading those at build time from their Git repo (see discussion in the TF-M link earlier here).

Downloading code from a non-Zephyr repo goes against Zephyr project policy, however, but NXP felt strongly about avoiding duplication of code, so the compromise we came up with was a default download flag that would be overridden to point to a different location for Zephyr to avoid the need to download anything, and also keep all dependencies inside the Zephyr github org. It leads to extra complications with Zephyr, but it was the acceptable compromise in this case.

It looks like the Zephyr fork of the NXP SDK should now be up to date, though. Is the NXP build still failing pointing to the latest version of https://github.com/zephyrproject-rtos/hal_nxp ? If that's the case, it seems more like something that should be filed as an issue with the Zephyr SDK.

joerchan:
> It looks like the Zephyr fork of the NXP SDK should now be up to date, though. Is the NXP build still failing pointing to the latest version of https://github.com/zephyrproject-rtos/hal_nxp ? If that's the case, it seems more like something that should be filed as an issue with the NXP SDK.

No, it appears it is actually not pointing to hal_nxp at all, instead it points to the NXP SDK files that are commited to this fork.
```
    # Supply path to NXP HAL sources used for TF-M build
    set(TFM_PLATFORM_NXP_HAL_FILE_PATH ${ZEPHYR_TRUSTED_FIRMWARE_M_MODULE_DIR}/platform/ext/target/nxp/)
```

So I suggest we clean that up afterwards.

SebastianBoe:
> We need to set TFM_PLATFORM_NXP_HAL_FILE_PATH to point to https://github.com/zephyrproject-rtos/hal_nxp (which is at 2.10.0 as of this comment), overriding the default download action.

Sorry Kevin, I don't understand the -1.

As I understand the situation we are pointing to a 2.10.0 of hal_nxp.
We are not downloading.
The 2.10.0 resides in tf-m, because we patched it to be 2.10.0.

Is it the filepath that we are pointing to that you protest?

E.g. you want us to point to the Zephyr NXP SDK?

This was already brought up as a TODO when we first re-directed the download I saw in some PR review.

I don't see why continuing to follow prior-art, and not fixing this TODO, should block the upmerge.

It was an issue before the upmerge, and it will continue to be an issue after the upmerge.

microbuilder (approved):

microbuilder:
@SebastianBoe Fair enough. I agree the NXP issue can and should be sorted out separately.

microbuilder:
@SebastianBoe Requires a manual rebase to be able to merge

SebastianBoe:
weird.

this branch is direct descendent of 7c887819, which is the master branch of https://github.com/zephyrproject-rtos/trusted-firmware-m.

Not sure what I'm supposed to be rebasing onto.

SebastianBoe:
I tried to rebase this branch onto https://github.com/zephyrproject-rtos/trusted-firmware-m.

But that is weird to me, to take two branches. Merge them together. And then rebase the resulting branch onto one of the two branches again.

Not sure if the rebase will produce anything that makes sense.

microbuilder:
> weird.
> 
> this branch is direct descendent of [7c88781](https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/7c88781953ce7146878bd8bd724e175fdbc298c5), which is the master branch of https://github.com/zephyrproject-rtos/trusted-firmware-m.
> 
> Not sure what I'm supposed to be rebasing onto.

Not sure at first glance either what the conflict is

SebastianBoe:
Is there a process for upmerging zephyr forks, or can we do whatever makes sense to us?

If we can do whatever we want I would suggest force-pushing this branch to master.

Another alternative would be to re-write the history so that it's tf-m-1.5, and then the three patches I mention in the PR description.

Merging two branches and then rebasing the result is not a viable option I think.

SebastianBoe:
@microbuilder : After thinking more about it, I believe a CLI git push from merge_tf-m-1.5 to zephyrproject-rtos:master is the most correct thing to do, but I would appreciate if you could confirm this.

microbuilder:
@SebastianBoe That will get the changes into master, but I'm not sure I understand why the PR here is refusing to do that for us. I don't see the conflict. I'm OK with doing a (normal) push to master, though, since the history is intact so there is no significant risk.

SebastianBoe:
git push.

thanks

SebastianBoe:
git push.

thanks

PR #64, opened 2021-12-21
platform: nucleo_l552ze_q: Reduce the size of the SPE partition

SebastianBoe:
The hex file zephyr_ns_signed.hex overlaps with something else at
0xC040E50 unless this patch, which reduces the size of the secure
partition, is applied.

When flash was allocated for OTP the intention was to maintain the
same flash usage by reducing another partition.

AFAICT from db07170a34ffe6e17bc68294bac37091023bdee1, the intention
was to remove an "unused" 4kB area.

But I cannot see that this was actually successful.

It is a bit confusing because the documented offsets appear to be
wrong. e.g. the secure image is documented in flash_layout.h to be
0x0001_4000.

But the map file tfm_s.map reports:

Memory Configuration

Name             Origin             Length             Attributes
FLASH            0x0000000011000400 0x000000000005f0c0 xr
RAM              0x0000000030000000 0x0000000000080000 xrw
VENEERS          0x000000001105f4c0 0x0000000000000340 xr
*default*        0x0000000000000000 0xffffffffffffffff

So 0x11000400, not 0x0001_4000.

There is likely a more correct solution, but this is a functional
hotfix for now.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/platform/ext/target/stm/nucleo_l552ze_q/include/flash_layout.h
+++ b/platform/ext/target/stm/nucleo_l552ze_q/include/flash_layout.h
@@ -115,7 +115,7 @@
 #define FLASH_ITS_AREA_OFFSET           (FLASH_PS_AREA_OFFSET+FLASH_PS_AREA_SIZE)
 #define FLASH_ITS_AREA_SIZE             (0x2000)   /* 8 KB */
 
-#define FLASH_S_PARTITION_SIZE          (0x2D000) /* S partition */
+#define FLASH_S_PARTITION_SIZE          (0x2C000) /* S partition */
 #define FLASH_NS_PARTITION_SIZE         (0x9000) /* NS partition */
 #define FLASH_PARTITION_SIZE (FLASH_S_PARTITION_SIZE+FLASH_NS_PARTITION_SIZE)
 /* Secure image primary slot */


microbuilder:
@erwango Any thoughts on this, since it's blocking us from upmerging to TF-M 1.5.0?

erwango:
@SebastianBoe One question to start, have you opened the point on TFM side ?

SebastianBoe:
Not yet.

I don't understand the problem very well, I'd like to be sure that we aren't doing anything wrong on the Zephyr side before taking it upstream.

SebastianBoe:
@erwango : I have now reported this upstream[0].

Unfortunately, the same hotfix does not work upstream. So I can't ask them to comment on the fix.

I would appreciate it if we could merge this hotfix now, and then re-work it later depending on what upstream finds.

[0] https://lists.trustedfirmware.org/pipermail/tf-m/2021-December/thread.html

microbuilder:
@SebastianBoe I understand the desire to get this in ASAP, but I think we should try to understand the problem here and check if there is a conflict between TF-M and Zephyr for this board that can/should be resolved, and that our assumptions in this PR match what ST intends. I didn't really understand this reponse either: https://lists.trustedfirmware.org/pipermail/tf-m/2021-December/002098.html

@erwango Who is responsible for the TF-M ports in general in ST? I know it's holiday season so they're likely out, but if the maintainer could have a look at this and confirm what the flash settings should be, it makes more sense to get this right now rather than just sweeping things under the carpet.

If this PR is 'right', I'm happy to approve and merge it in, but I'd like to know first that we're doing the right thing.

erwango:
@microbuilder TFM maintainer is @jamike. He's having a look right now, we should have an answer by tomorrow.

erwango:
@SebastianBoe @microbuilder, I think we found the issue. I'm posting a PR in Zephyr repo right now.
Though, I can't build using TFM 1.5 due to the following error:
```
Traceback (most recent call last):
  File "/local/mcu/zephyrproject/modules/tee/tf-m/trusted-firmware-m/bl2/ext/mcuboot/scripts/wrapper/wrapper.py", line 21, in <module>
    import imgtool.main
  File "/local/mcu/zephyrproject/bootloader/mcuboot/scripts/imgtool/main.py", line 23, in <module>
    import imgtool.keys as keys
  File "/local/mcu/zephyrproject/bootloader/mcuboot/scripts/imgtool/keys/__init__.py", line 27, in <module>
    from cryptography.hazmat.primitives.asymmetric.ed25519 import (
ModuleNotFoundError: No module named 'cryptography.hazmat.primitives.asymmetric.ed25519'
[603/610] Building C object app/CMakeFiles/tfm_api...sted-firmware-m/interface/src/tfm_crypto_ipc_api.o
ninja: build stopped: subcommand failed.

```
My pip install is up to date vs `'cryptography` module and mcuboot requirements (v2.9 vs v2.6).
Is there something else that should be done for this ?

erwango:
The zephyr PR which should fix the issue https://github.com/zephyrproject-rtos/zephyr/pull/41432

SebastianBoe:
Closing this in favour of 41432

PR #65, opened 2022-01-04
Nrf patches

SebastianBoe:
Various nrf patches that were applied after TF-M 1.5, but are importantant enough that
they can't wait for a new upmerge.


microbuilder (approved):
Tested on `nrf5340dk_nrf5340_cpuapp_ns` ... I don't have an nRF91 on hand to test with, but presumably you've done the tests there locally?

joerchan (approved):

PR #67, opened 2022-03-06
scripts: Allow use of bash outside of /bin

otavio:
Even though bash is commonly available as /bin/bash there are
exceptions (e.g NixOS). This commit allow the use of the scripts in my
environment and is generic.

Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>


cfriedt (approved):

stephanosio (commented):
Is this script actually called by the Zephyr build system? If so, there should be a Zephyr-side PR that updates `west.yml`.

Also this file is from upstream TF-M, so this patch should be submitted upstream as well.

otavio:
It is; I spot this running some of our tests.

platform/ext/target/stm/common/scripts/postbuild.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -
+#!/usr/bin/env bash
```
d3zd3z, line 1: I'm fine with this, although do we know if the included script uses any bash specific features? Otherwise just `#!/bin/sh` could be fine.

otavio: I didn't check it; it could be checked but either way it'd be another patch.

otavio: No; I did this here directly.

joerchan: Can you please submit this PR to the TF-M upstream repository?
We can cherry-pick it to the zephyr fork once it is merged upstream.

Have a look here for how to do it:
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/tfm/docs/contributing/contributing_process.html

joerchan (changes requested):
Should be submitted upstream first.

Cherry-pick to downstream using `git cherry-pick -x`.

Then needs a manifest update, the zephyr west.yml file.

microbuilder:
We're planning on doing an upmerge next week to TF-M, so this should get pulled in during the upmerge to the tip of TF-M.

PR #70, opened 2022-04-04
Update to TFM 1.6.0

theotherjimmy:
Merge remote-tracking branch 'tforg/1.6.0'


microbuilder:
@joerchan @d3zd3z For testing purposes so that we can take advantage of some post 1.5.0 commits around MbedTLS, etc.

theotherjimmy:
Yep. Hence the WIP PR.

joerchan:
Thanks for the heads-up. Is 500+ commits right?

joerchan:
> > Thanks for the heads-up. Is 500+ commits right?
> 
> We can squash this to a single commit as well. The history does still exist upstream to consult commit-by-commit.

Ignore, that, I forgot which repository I was looking at.
No we shouldn't squash anything.

theotherjimmy:
Since it's sort of impossible to review: I removed the revert of the NXP HAL addition to tfm.

joerchan:
> Since it's sort of impossible to review: I removed the revert of the NXP HAL addition to tfm.

Maybe we should address this issue separately first?
Could we get the NXP maintainers to fix this problem, we shouldn't have an SDK committed into the repository in this way.

microbuilder:
@joerchan We used to have the NXP SDK files upstream as part of TF-M like other platforms do, but NXP made the decision to remove them and download the SDK at compile time via cmake. This conflicts with Zephyr's project policies, however, that we can't dynamically download dependencies at compile time, and everything needs to be inside the Zephyr github org, which is how we ended up with the macro being defined here:

- https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10225

And it's implementation in Zephyr here:

- https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/trusted-firmware-m/CMakeLists.txt#L218=

What should likely be happening, however, is that we should be pointing to Zephyr's fork of the NXP HAL directly (https://github.com/zephyrproject-rtos/hal_nxp), though, which addresses the project management requirements from Zephyr.

@theotherjimmy Can you poke around at this and see what kind of maintainable solution makes sense, changing the directory we point to with `TFM_PLATFORM_NXP_HAL_FILE_PATH` to see if you can get it to play well with Zephyr's fork of the NXP SDK?

joerchan:
@microbuilder Yes, I agree that pointing to NXP HAL directly is the correct solution. We have done the same thing for HAL Nordic.
You can see it added here: https://github.com/zephyrproject-rtos/zephyr/commit/5754ec4b2145e158ad57ba46f9ace96e7d389136
And the TF-M change: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/13079

We took a quick look at NXP to see if we could update it, but it wasn't matching. Maybe not the right version or something I don't really remember, but since it wasn't trivial to fix we didn't spend more time on it.

So I'm wondering if someone from NXP could spend some time to fix this for their platform.

d3zd3z:
> We can squash this to a single commit as well. The history does still exist upstream to consult commit-by-commit.

Definitely not. This should be a merge with the upstream so that we have the full history, with the exact upstream commits.

joerchan:
What's up with the revert?

microbuilder:
This is causing failures with the FWU service, and is the current blocker to be able to merge 1.6.0 support into Zephyr.

joerchan:
Thank you for the explanation. Just curious as to what is happening.

joerchan:
Need to fix the static version: `Booting TF-M v1.6.0-RC3+31d4dce6`

microbuilder (approved):

microbuilder:
> Need to fix the static version: `Booting TF-M v1.6.0-RC3+31d4dce6`

Should be done.

carlescufi (approved):

PR #72, opened 2022-05-19
TF-M: flash: Add missing 304KB of U585I target flash memory

rajkan01:
U585i target flash partition total allocated size is 1755 KB, the total available flash of 2MB and nearly 304KB unused. this PR update the flash partition to use that extra unused flash memory to add to the secure partition size, which helps for some complex examples like ML stack fit to secure side as it requires more memory.

with update creates a new flash layout as below:

primary slot:
- Secure partition - 384 KB
- Non-Secure - 512 KB

Secondary slot:
- Secure partition - 384 KB
- Non-Secure - 512 KB

Signed-off-by: Rajkumar Kanagaraj <rajkumar.kanagaraj@linaro.org>


microbuilder:
See: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/15270

microbuilder:
Will be resolved next upmerge.

PR #74, opened 2022-08-11
platform: musca_s1: only declare tfm_ns if NS is enabled

fabiobaltieri:
Make the last target_sources(tfm_ns...) conditional to NS=TRUE. The
current setup is causing bulid issues for musca_s1 since:

c51505f661 modules: tfm: Exclude non-secure TF-M application from build

Fails with:

CMake Error at platform/ext/target/arm/musca_s1/CMakeLists.txt:179:
  Cannot specify sources for target "tfm_ns" which is not built by this
  project.

### DIFF
--- a/platform/ext/target/arm/musca_s1/CMakeLists.txt
+++ b/platform/ext/target/arm/musca_s1/CMakeLists.txt
@@ -171,14 +171,16 @@ target_include_directories(platform_ns
         ${CMAKE_SOURCE_DIR}/platform/ext/cmsis
 )
 
-# Adding new defined handler function to the library where the weak handler
-# funtion symbol are, to override weak handler funtion.
-# Otherwise the weak handler funtion cannot be replaced by the new defined
-# hanlder function which symbol is in another library.
-target_sources(tfm_ns
-    PRIVATE
-        $<$<BOOL:${TEST_NS_FPU}>:${CMAKE_CURRENT_SOURCE_DIR}/ns_fp_test_interrupt.c>
-)
+if(NS)
+    # Adding new defined handler function to the library where the weak handler
+    # funtion symbol are, to override weak handler funtion.
+    # Otherwise the weak handler funtion cannot be replaced by the new defined
+    # hanlder function which symbol is in another library.
+    target_sources(tfm_ns
+        PRIVATE
+            $<$<BOOL:${TEST_NS_FPU}>:${CMAKE_CURRENT_SOURCE_DIR}/ns_fp_test_interrupt.c>
+    )
+endif()
 
 #========================= Platform BL2 =======================================#
 


microbuilder:
@fabiobaltieri Do you mind proposing this as a change upstream to TF-M first? As is, this will have to be re-applied every time we update TF-M.

If you're not familiar with the TF-M submission process, let me know and I can put together a change request myself.

microbuilder:
This seems to have already been [removed upstream](https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/commit/platform/ext/target/arm/musca_s1/CMakeLists.txt?id=f0ef274261e332e0586d3b013cd4100e52c219a0), combined with some other changes. As such, I'm OK with merging this since it won't cause a conflict during the next upmerge.

Any objections @joerchan ?

joerchan:
> This seems to have already been [removed upstream](https://git.trustedfirmware.org/TF-M/trusted-firmware-m.git/commit/platform/ext/target/arm/musca_s1/CMakeLists.txt?id=f0ef274261e332e0586d3b013cd4100e52c219a0), combined with some other changes. As such, I'm OK with merging this since it won't cause a conflict during the next upmerge.
> 
> Any objections @joerchan ?

No objections, but this commit should then be reverted before the next upmerge then.
Could add a note in the commit message for that.

fabiobaltieri:
Hey when is the next upmerge planned to happen? May as well wait for that and save you some back and forth...

joerchan:
> Hey when is the next upmerge planned to happen? May as well wait for that and save you some back and forth...

At the moment we have only upmerged the releases, and TF-M has a release cadence of 6 months. So we shouldn't wait for the upmerge for this fix.

joerchan:
@fabiobaltieri Can you please update the commit message. We need a signed-off for the commit message. (`git commit --amend -s`).

I suggest adding a note: "Temporary fix for zephyr fork of trusted-firmware-m, this commit is not needed in upstream TF-M. This commit should be reverted in the next upmerge"

fabiobaltieri:
> I suggest adding a note: "Temporary fix for zephyr fork of trusted-firmware-m, this commit is not needed in upstream TF-M. This commit should be reverted in the next upmerge"

Cool, done.

joerchan (approved):

microbuilder (approved):

microbuilder:
@fabiobaltieri Can you open up a PR in Zephyr updating the manifest to point to this commit?

fabiobaltieri:
> @fabiobaltieri Can you open up a PR in Zephyr updating the manifest to point to this commit?

Yeah I'll add it to https://github.com/zephyrproject-rtos/zephyr/pull/48915, have few other unrelated breakages to sort out.

PR #77, opened 2022-09-30
build: gcc: Force DWARF v4

nordicjm:
This forces DWARF version 4 output so that zephyr debugging and usage still works with the pyelftools library which does not currently support v5.

### DIFF
--- a/toolchain_GNUARM.cmake
+++ b/toolchain_GNUARM.cmake
@@ -49,6 +49,8 @@ macro(tfm_toolchain_reset_compiler_flags)
         -mthumb
         -nostdlib
         -std=c99
+        # Force DWARF version 4 for zephyr as pyelftools does not support version 5 at present
+        -gdwarf-4
         $<$<OR:$<BOOL:${TFM_DEBUG_SYMBOLS}>,$<BOOL:${TFM_CODE_COVERAGE}>>:-g>
     )
 endmacro()


stephanosio:
> @nordicjm I would prefer to send this directly to TF-M first, and then cherry-pick the change once approved there.

But this is more or less Zephyr-specific issue (pyelftools issue, which is used by Zephyr `tfm_ram_report` implementation). Does it make sense for this to be upstreamed/will they accept it ?

nordicjm:
The reason I submitted this here rather than upstream is zephyr 3.2 is going to be released in a few hours, and it was to add a quick fix rather than go through testing and approval via TF-M.

joerchan:
@nordicjm Still, could you open the PR, add a comment to the upstream PR in the commit message?

I'm fine with not having it approved in upstream, but I would prefer to be able to see what the status is of the downstream patch.

Need to update the SHA entry for TF-M in zephyr as well.

SebastianBoe (approved):

microbuilder (approved):
Approving, but I'd like to at least see us try to get this upstream to reduce the maintenance burden on us longer term.

stephanosio (approved):

nordicjm:
@microbuilder who can I contact about getting my TF account reset? Doesn't seem to like that it's been renamed and I can't change the github name from the TF settings, won't let me get git hook files or submit patches

joerchan (approved):

microbuilder:
> @microbuilder who can I contact about getting my TF account reset? Doesn't seem to like that it's been renamed and I can't change the github name from the TF settings, won't let me get git hook files or submit patches

Reply via discord, to avoid dropping emails here

PR #78, opened 2022-10-05
SPM: Return PSA_ERROR_PROGRAMMER_ERROR for invalid parameters in NS c…

joerchan:
…alls

Currently S calls are able to return PSA_ERROR_PROGRAMMER_ERROR for malformed packet parameters, but NS calls will just stay in the error handler causing a halt or reboot of device, i.e. this is a very easy Denial-Of-Service attack.
Make sure that the SPM return the error code to NS in this case without halting the whole device.

This patch will not be sent upstream since the library model is being deprecated and fixes are no longer accepted.

Signed-off-by: Antonio de Angelis <Antonio.deAngelis@arm.com>
Change-Id: I557ce35ccce0354421ccd9b2140684a1d833bf0e
Signed-off-by: Joakim Andersson <joakim.andersson@nordicsemi.no>


microbuilder:
Looks sensible to me, thanks @joerchan, but I've asked @d3zd3z to have a look as well.

microbuilder (approved):

PR #80, opened 2022-11-02
build: gcc: Masking mbedtls warning

ceolin:
mbedTLS upstream code contains a warning about const variable set but unused. Just ignore it to avoid Zephyr tests to fail.

zephyr/modules/crypto/mbedtls/library/aes.c:307:23: warning: 'RT0' defined but not used [-Wunused-const-variable=]
      307 | static const uint32_t RT0[256] = { RT };
          |                       ^~~
zephyr/modules/crypto/mbedtls/library/aes.c:200:28: warning:
'RSb' defined but not used [-Wunused-const-variable=]
      200 | static const unsigned char RSb[256] =

Fixes #51025 (on Zephyr)

Signed-off-by: Flavio Ceolin <flavio.ceolin@intel.com>

### DIFF
--- a/toolchain_GNUARM.cmake
+++ b/toolchain_GNUARM.cmake
@@ -39,6 +39,7 @@ macro(tfm_toolchain_reset_compiler_flags)
         -Wall
         -Wno-format
         -Wno-return-type
+        -Wno-unused-const-variable
         -Wno-unused-but-set-variable
         -c
         -fdata-sections


d3zd3z (approved):

microbuilder:
@ceolin This should really go upstream first. If it's blocking things, I'm OK with merging, but we really need to push upstream as well to avoid problems during the next update to TF-M.

microbuilder (approved):

microbuilder:
Requires cherry picking if/when merged: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/17599

PR #88, opened 2023-03-22
TF-M: Replace zephyr 3.3.0 fixes with official fixed from upstream TF-M

joerchan:


config/config_base.cmake
```diff
@@ -121,6 +121,7 @@ set(TFM_PARTITION_PLATFORM              OFF         CACHE BOOL      "Enable Plat
 ################################## Dependencies ################################
 
 set(MBEDCRYPTO_PATH                     "DOWNLOAD"  CACHE PATH      "Path to Mbed Crypto (or DOWNLOAD to fetch automatically")
+set(MBEDCRYPTO_FORCE_PATCH              OFF         CACHE BOOL      "Always apply MBed Crypto patches")
```
frkv, line 124: Mbed

joerchan: Don't want to change the already accepted TF-M patch. It can be changed in TF-M directly and does not require backporting.

microbuilder (approved):

PR #91, opened 2023-04-27
platform: nordic_nrf: Add nrf9161 DK support

joerchan:


SebastianBoe (approved):
All of this has been posted for review upstream and that is where the review should be done so this is +1.

microbuilder (approved):

PR #92, opened 2023-05-16
Patchset: Align TFM to mbedTLS

rajkan01:
## Reason for this patch set

Since an update to TFM 1.8 is only planned after the Zephyr 3.4.0 release is complete, the ARM TF-M team recommended the below patch set be applied on top of TF-M 1.7.0. These cherry-picks mitigate issues caused by the most recent version of mbedTLS being updated to use PSA APIs, causing conflicts when using mbedTLS with TF-M.

See: https://review.trustedfirmware.org/q/topic:%22psa_crypto_headers_adjustments%22+(status:open%20OR%20status:merged)

## Summary of changes

- Define the type `psa_key_attributes_t` in the same way as Mbed TLS, as a `struct psa_key_attributes_s` which has different definitions on the client and server sides. This avoids needing to patch upstream headers.
- Reduce divergence between PSA Crypto headers available in TF-M and Mbed TLS. 
- Consistently use `mbedtls_svc_key_id_t` for `psa_key_id_t` in the headers 
- Concentrate IMPDEF behaviour in crypto_struct and crypto_platform
- Make sure the same config file is used for client and server views, i.e. make sure that ENCODES_OWNER config define is passed in a dedicated compile switch privately to the crypto service
- Add `\emptydescription` for Doxygen when required
- clean up the `psa_crypto_config` cmake target
- Reduces the degrees of freedom when including PSA crypto headers. The PSA Crypto headers are always guaranteed to be in the same directory of the main `psa/crypto.h`.
- Header file `crypto_platform.h` is included by `crypto.h`, which is installed when Crypto service is enabled. So `crypto_platform.h` also needs to be installed.

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -78,6 +78,7 @@ if (TFM_PARTITION_CRYPTO)
                         ${INTERFACE_INC_DIR}/psa/crypto_compat.h
                         ${INTERFACE_INC_DIR}/psa/crypto.h
                         ${INTERFACE_INC_DIR}/psa/crypto_client_struct.h
+                        ${INTERFACE_INC_DIR}/psa/crypto_platform.h
                         ${INTERFACE_INC_DIR}/psa/crypto_sizes.h
                         ${INTERFACE_INC_DIR}/psa/crypto_struct.h
                         ${INTERFACE_INC_DIR}/psa/crypto_types.h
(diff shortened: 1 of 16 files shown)


microbuilder:
@joerchan Have time to take a look at this one?

d3zd3z (approved):

PR #93, opened 2023-06-08
platform: stm32u5xx: Enable RTC-RTC_EX on platform_s

nandojve:
The current stm platform not have rtc functions available on platform_s. In this case, tamper and backup registers can are not available to users consume on a secure partition. This enable those features on platform_s as private content.

Change-Id: I8edad9ff441011fe68964c2b95b7dd0bc4630201


d3zd3z (approved):
Cherry pick from upstream.

nandojve:
Already in main : )

PR #94, opened 2023-07-10
tf-m 1.8.0

d3zd3z:
Upgrade TF-M to 1.8.0.


microbuilder (approved):
Odd name for the PR (Should perhaps be `TF-M 1.8.0` 😉), but passing all tests here.

microbuilder:
@joerchan Mind having a look now that the parent PR is passing in CI?

joerchan:
The commit "Build: Add stub functions of system calls" confuses me.

Normally in the commit messages we have either "cherry-picked from <x>" when taking a commit from TF-M master branch, or "Upstream PR <gerrit-link>" when a commit is still in review.
If neither of these are the case it would be helpful to have this in the commit message explaining why it is not from TF-M, or sent to TF-M.
The thing that bothers me about this commit is that it is referencing a TF-M tests repository commit ID that I cannot find. And I cannot see that any changes was made in the test-repository for this commit, so I am suspecting that we are missing a commit.

joerchan (approved):
Not clear what the test-update was, but if the tests pass I guess it should be OK.
The linked PR did not clarify where the test-update change was.

PR #95, opened 2023-07-31
Update to TF-M for LPC parts

d3zd3z:
Some updates to the SDK for the LPC55S69 got dropped by the upgrade to 1.8.0.  Re-apply these patches to bring the SDK back in.


microbuilder:
@dleach02 Is there someone on your team that can do some testing on this with the LPC55s69, just to get a +1 from someone at NXP?

dleach02:
I've forward this internally to get a reviewer

butok:
Hi David and Kevin,
Thank you for the NXP TFM support.

This commit adds the SDK drivers inside the TFM NXP port directory. So we can add +Vote.

But I would like to start a discussion for possible improvements:

1. **Right place for drivers.**
   The NXP TFM has 3 versions:
    - **MCUx SDK TFM**:  Drivers are placed outside of TFM folder, part of an SDK application project.
    - **Upstream TFM**: Drivers are loaded by CMAKE from the GitHub SDK to the TFM NXP folder.
     - **Zephyr TFM: Drivers are placed inside the TFM NXP port folder.**
 
   I am really new to Zephyr, but have feeling that the Zephyr TFM approach most probably adds a duplication.
   **The HW drivers should by outside of TFM folder added based on an application device tree. Right?
    How do you solve it for other platforms?**

2. **Driver version**
   Based on the commit name, drivers are taken from old SDKv2.10.
   The latest dev. Upstream TFM version is using SDK v2.13 drivers.

3. **Not all these drivers are used by TFM.**
   I need to review the driver list, and delete not used, at least for the Upstream TFM.

It should not block the commit.
Just needed to figure out the right solution for the issue (1).

joerchan (approved):

microbuilder (approved):

PR #100, opened 2023-10-04
platform: nordic_nrf: Set UART pins using pinctrl method

joerchan:
Use the pinctrl method to define the UART pins for the nordic platform
UART driver.
This makes it easier to assign the UART pins from devicetree information
which is used in out-of-tree board support.


carlescufi:
@anangl @oyvindronningstad please take a look

gmarull (approved):

platform/ext/target/lairdconnectivity/bl5340_dvk_cpuapp/RTE_Device.h
anangl, line 1: Changes in this file should not be done in commit "platform: Add platform default UART control definition" but rather in the "platform: nordic_nrf: Set UART pins using pinctrl method" one.

platform/ext/target/nordic_nrf/common/core/cmsis_drivers/Driver_USART.c
```diff
@@ -30,6 +33,43 @@
...
+    .baudrate = NRF_UARTE_BAUDRATE_115200,                                     \
+    .interrupt_priority = NRFX_UARTE_DEFAULT_CONFIG_IRQ_PRIORITY,              \
+    .config  = {                                                               \
+        .hwfc   = NRF_UARTE_HWFC_DISABLED,                                     \
+        .parity = NRF_UARTE_PARITY_EXCLUDED,                                   \
+        .stop   = NRF_UARTE_STOP_ONE,                                          \
+    },                                                                         \
+}
+
+void uart_config_set_uart_pins(nrfx_uarte_config_t *uart_config,
```
anangl, line 53:
```suggestion
static void uart_config_set_uart_pins(nrfx_uarte_config_t *uart_config,
```

joerchan: @anangl I don't disagree, but these patches are already merged in upstream TF-M now.
I will submit a patch upstream for this, but I think we can skip this change until next sync with TF-M.

anangl (approved):

PR #104, opened 2024-04-09
patch TF-M 2.0 to work with MbedTLS 3.6.0

tomi-font:
Patches provided by @adeaarm in https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/27718.


frkv (approved):
LGTM (But making a copy of legacy Mbed TLS include files needs to be discussed in upstream TF-M project)

Follow up: It has been discussed as a necessity right now, but it will be re-addressed when moving to TF-PSA-Crypto

ithinuel (approved):

ceolin:
@d3zd3z ping

d3zd3z (commented):
I'm not sure I understand what is happening here. Why are we pulling in what looks like a full copy of the headers from mbedtls instead of just using the mbedtls headers. This just gives opportunity for these headers to be out of sync and cause difficult to find problems.

frkv:
> I'm not sure I understand what is happening here. Why are we pulling in what looks like a full copy of the headers from mbedtls instead of just using the mbedtls headers. This just gives opportunity for these headers to be out of sync and cause difficult to find problems.

This has been discussed offline with @adeaarm (TF-M codeowner). This is done to ensure fewer dependencies when building TF-M. Note that the Mbed TLS headers are copied 1 to 1 and that this duplication is targeted for redesign once the self-contained TF-PSA-Crypto crypto toolbox repository is in place (timeline unknown). There is a strong push to stay with these releases as they are LTS releases and it is expected that any changes in them will be fully supported  (backporting) for both Mbed TLS 3.6.x and TF-M 2.1.x

adeaarm:
LGTM

ithinuel (approved):

PR #108, opened 2024-07-10
build: remove the warning when the version Git tag is not found

tomi-font:
Cherry picked from https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/29897.


ithinuel (commented):
Waiting on the TF-M resolution to see if/how they want it to fix that before providing an approval on this PR. 👀

Vge0rge (approved):

ithinuel (approved):

PR #109, opened 2024-07-19
platform: ext: add local copy of CMSIS v6 files

tomi-font:
This adds all the files (minus `.git*` and `CMSIS/Documentation/` for saving on size) from the CMSIS v6 repository
(https://github.com/ARM-software/CMSIS_6) at the revision `d0c460c169` as defined in `lib/ext/cmsis/CMakeLists.txt`.
The patch `lib/ext/cmsis/0001-iar-Add-missing-v8.1m-check` is applied on top.

This is because as of v2.1.0 TF-M has updated to CMSIS v6 and switched from hosting the sources to depending on the upstream repository, cloning it at build time.

To prevent a download from happening during the build, CMSIS v6 sources are pushed and the CMSIS_PATH CMake variable is used to point to them.

### DIFF
--- a/platform/ext/cmsis/.devcontainer/ubuntu-22.04/Dockerfile
+++ b/platform/ext/cmsis/.devcontainer/ubuntu-22.04/Dockerfile
@@ -1,0 +1,59 @@
+FROM  --platform=linux/amd64 ubuntu:22.04
+
+SHELL ["/bin/bash", "-c"]
+
+ARG USERNAME=user
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+ARG DEBIAN_FRONTEND=noninteractive 
+
+RUN apt-get update && \
+    apt-get -y install \
+        build-essential \
+        curl \
+        gdb \
+        gh \
+        git \
+        less \
+        libncurses5 \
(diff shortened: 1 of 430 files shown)


Vge0rge (approved):

valeriosetti (approved):

tomi-font:
> It doesn't seem to be an appropriate answer for the mid/long term but considering it's criticality for the release I'm giving it a ✅

Yeah, it can then be reverted and the proper solution implemented.

d3zd3z (changes requested):
Two concerns here:
- Why are we pulling this into trusted-firmware-m instead of creating a module for cmsis 6? We're already pulling in cmsis_5.
- We need approval from the TSC in order to pull in additional external code.

tomi-font:
> Why are we pulling this into trusted-firmware-m instead of creating a module for cmsis 6? We're already pulling in cmsis_5.

Because this is a quick fix that was suggested to get it in 3.7. Indeed creating a module is the proper way and shall be done. I don't know if people will accept creating a new module directly and integrating it into 3.7, which is why this quick fix.

d3zd3z (approved):
No objections from the TSC

PR #110, opened 2024-08-19
platform: ext: nordic_nrf: align pinctrl with v3.7-branch

gmarull:
- Remove redundant bits
- Align function and pin fields position/mask.

cc: @JordanYates

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/common/nrf-pinctrl.h
+++ b/platform/ext/target/nordic_nrf/common/core/common/nrf-pinctrl.h
@@ -19,35 +19,21 @@
  * - 6..0:   Pin number (combination of port and pin).
  */
 
+/* NOTE: Keep in sync with Zephyr's nrf-pinctrl.h */
+
 /**
  * @name nRF pin configuration bit field positions and masks.
  * @{
  */
 
 /** Position of the function field. */
-#define NRF_FUN_POS 16U
+#define NRF_FUN_POS 17U
 /** Mask for the function field. */
-#define NRF_FUN_MSK 0xFFFFU
-/** Position of the invert field. */
-#define NRF_INVERT_POS 14U
-/** Mask for the invert field. */
-#define NRF_INVERT_MSK 0x1U
-/** Position of the low power field. */
-#define NRF_LP_POS 13U
-/** Mask for the low power field. */
-#define NRF_LP_MSK 0x1U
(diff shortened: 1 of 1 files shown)


Vge0rge:
@joerchan  is right @gmarull, you need to a PR in the upstream repo:
https://git.trustedfirmware.org/plugins/gitiles//TF-M/trusted-firmware-m.git/

And then bring the cherry pick here. There are exceptions to this rule when there is no other way but this PR doesn't seem to be one of them. 

Also keep in mind that @joerchan is not working for Nordic anymore and he is not responsible for reviewing PRs here.

gmarull:
> @joerchan is right @gmarull, you need to a PR in the upstream repo: https://git.trustedfirmware.org/plugins/gitiles//TF-M/trusted-firmware-m.git/
> 
> And then bring the cherry pick here. There are exceptions to this rule when there is no other way but this PR doesn't seem to be one of them.
> 
> Also keep in mind that @joerchan is not working for Nordic anymore and he is not responsible for reviewing PRs here.

Who can take care of this? I've never contributed to TF-M directly, and I guess someone else will make this job quicker.

Vge0rge:
I can have a look on this hopefully tomorrow.

Vge0rge:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/30839

You can cherry pick this commit here as fromlist if you want it fast, or wait until it is merged to get it as fromtree.

gmarull:
@Vge0rge should be ready

gmarull:
It looks like stuff has been merged to `main`, so we may have to branch tf-m to fix v3.7 release, it depends if changes in main require changes in Zephyr as well...

Vge0rge:
You can actually remove the fromtree completely now, we do that in our nrf fork but here in zephyr as long as you cherry pick the commit it seems good enough. 

I will be out for a couple of weeks starting from Tuesday though so please advice @tomi-font on how to proceed with this.

Vge0rge (approved):

tomi-font:
It looks like we may need to branch for 3.7 because https://github.com/zephyrproject-rtos/zephyr/pull/77783 just brought some upstream cherry picks in this repository to make Zephyr's TF-M compatible with the newer version of MCUboot. So merging this PR and making 3.7 point to the head of this repository won't work unless folks want to update MCUboot there as well.
@carlescufi

PR #117, opened 2024-11-25
Increase the size of heap memory of the crypto partition to allow RSA 2048-bits signatures

valeriosetti:
This PR cherry-picks:

- https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/33552
- https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/33523

from the upstream TF-M project. The goal is to fix heap memory size for crypto partition (basically Mbed TLS) in order to allow RSA signatures up to 2048-bits (previously only RSA with 1024-bits key size were possible).

At the same time stack size for the crypto partition is reduced because:
- the previous value was too large;
- mitigate the effects of the heap increase.

This fix is required to resolve: https://github.com/zephyrproject-rtos/zephyr/issues/79864

### DIFF
--- a/config/config_base.h
+++ b/config/config_base.h
@@ -123,7 +123,7 @@
 
 /* The stack size of the Crypto Secure Partition */
 #ifndef CRYPTO_STACK_SIZE
-#define CRYPTO_STACK_SIZE                      0x1B00
+#define CRYPTO_STACK_SIZE                      0x1800
 #endif
 
 /* FWU Partition Configs */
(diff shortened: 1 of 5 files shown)


tomi-font (changes requested):
Please rename this PR to something more explicit (we know that the commits are cherry picks by looking at them, but what is this PR for?) and cherry pick the commits with the `-x` option.

tomi-font (approved):

ithinuel (approved):

d3zd3z (approved):

PR #118, opened 2024-12-23
Remove Ethos patches

wearyzen:
What is changed?

- Cherry-picked below patches from TF-M to make the code compatible with the latest Ethos-U driver.
  - https://github.com/TrustedFirmware-M/trusted-firmware-m/commit/5a92b9ad66c313efa73d462adec4067af81902a9
  - https://github.com/TrustedFirmware-M/trusted-firmware-m/commit/75c0999bd55b64b260529651e7ca994989508e1b

Note that the corstone320 changes from the original commits are skipped from the cherry pick since the support for corstone320 is not yet available in Zephyr.

Why is this change required?
- This is to be inline with Zephyr's rules to not fetch code from external repo and use Zephyr's locally available copy of Ethos-U driver.

Zephyr PR to test this : https://github.com/zephyrproject-rtos/zephyr/pull/83360


ithinuel (approved):

tomi-font (changes requested):
- The two commits are not identical in diff compared to their upstream counterparts. Please make them clean cherry picks.
- Also, please cherry pick them with the `-x` flag so that we know from the commit message that those commits are cherry picks and from which upstream commits.

wearyzen:
> * The two commits are not identical in diff compared to their upstream counterparts. Please make them clean cherry picks.
> * Also, please cherry pick them with the `-x` flag so that we know from the commit message that those commits are cherry picks and from which upstream commits.

Hi @tomi-font, thank you for the review. I have updated the commit message and the PR description to reflect that corstone320 changes are skipped from the cherry-pick.

tomi-font:
> Hi @tomi-font, thank you for the review. I have updated the commit message and the PR description to reflect that corstone320 changes are skipped from the cherry-pick.

What is the reason for skipping those changes? Is the skipping needed?

wearyzen:
> > Hi @tomi-font, thank you for the review. I have updated the commit message and the PR description to reflect that corstone320 changes are skipped from the cherry-pick.
> 
> What is the reason for skipping those changes? Is the skipping needed?

The first commit "npu: Update NPU driver to version 24.08" expects that `platform/ext/target/arm/mps4/corstone320/target_cfg.c` is already present but since corstone320 support is not available the cherry-pick isn't a clean and it needs more commits to bring int which becomes a big tasks that it has to be. I think it would make sense to bring those changes when the support for corstone320 is added or when the tfm is updated to the latest version.

tomi-font (approved):
I see. Makes sense. Thanks for the expanation.

PR #119, opened 2025-01-30
module: add security info to module.yml

mrodgers-witekio:
Add CPE and PURL references to module.yml file for use by Zephyr's SPDX generation tool.

See github issue in Zephyr repo: https://github.com/zephyrproject-rtos/zephyr/issues/53479


zephyr/module.yml
```diff
@@ -3,3 +3,7 @@ name: trusted-firmware-m
 build:
   cmake-ext: True
   kconfig-ext: True
+security:
+  external-references:
+    - cpe:2.3:o:arm:trusted_firmware-m:2.1.1:-:*:*:*:*:*:*
+    - pkg:github/TrustedFirmware-M/trusted-firmware-m@2.1.1
```
kartben, line 9:
```suggestion
    - pkg:github/TrustedFirmware-M/trusted-firmware-m@TF-Mv2.1.1
```

kartben (changes requested):

zephyr/module.yml
kartben: for github package it wants to be a SHA or tag, I believe

mrodgers-witekio: Good spot, I just put in the same version as in the CPE, but for the `github` type specifically this seems to want a commit or tag: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst

I'll go check the other PRs for the same thing as well

kartben: @mrodgers-witekio I stumbled upon https://github.com/zephyrproject-rtos/hostap/pull/77 and wasn't aware of that syntax for generic VCS. Wonder if it would be better to employ it here, given that github is not the canonical repo for tf-m / tf-a and rather a read-only mirror of repos hosted at https://git.trustedfirmware.org/plugins/gitiles/
I guess it's unlikely the mirrors will stop being updated, but git.trustedfirmware.org is probably the reference nevertheless. Thoughts?

mrodgers-witekio: To be honest I only spotted the "generic VCS" option after this PR was created. I think you're right that the trustedfirmware.org repo is probably better

kartben (approved):

tomi-font (approved):

PR #120, opened 2025-02-14
Boot: Add flash_area_get_sector function

de-nordic:
The commit adds flash_area_get_sector function that is used for obtaining parameters of a sector, at a given offset, from flash area.

### DIFF
--- a/bl2/src/flash_map.c
+++ b/bl2/src/flash_map.c
@@ -377,6 +377,26 @@ int flash_area_write(const struct flash_area *area, uint32_t off,
     return 0;
 }
 
+int flash_area_get_sector(const struct flash_area *fa, uint32_t off,
+                          struct flash_sector *sector)
+{
+    if (fa == NULL || sector == NULL) {
+        return -1;
+    }
+
+    if (off >= fa->fa_size) {
+        return -1;
+    }
+
+    sector->fs_off = off & ~(FLASH_AREA_IMAGE_SECTOR_SIZE - 1);
+    sector->fs_size = FLASH_AREA_IMAGE_SECTOR_SIZE;
+
+    BOOT_LOG_DBG("area %d: offset=0x%x to sector=0x%x, size=0x%x, length=0x%x",
+                 fa->fa_id, off, sector->fs_off, sector->fs_size);
+
+    return 0;
+}
+
 int flash_area_erase(const struct flash_area *area, uint32_t off, uint32_t len)
 {
     ARM_FLASH_INFO *flash_info;
--- a/bl2/ext/mcuboot/include/flash_map/flash_map.h
+++ b/bl2/ext/mcuboot/include/flash_map/flash_map.h
@@ -160,6 +160,11 @@ int flash_area_erase(const struct flash_area *area, uint32_t off, uint32_t len);
 uint32_t flash_area_align(const struct flash_area *area);
 
 /*
+ * Return sector parameters at specified offset of flash area,
+ */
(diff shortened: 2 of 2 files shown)


bl2/src/flash_map.c
```diff
@@ -377,6 +377,26 @@ int flash_area_write(const struct flash_area *area, uint32_t off,
...
+{
+    if (fa == NULL || sector == NULL) {
+        return -1;
+    }
+
+    if (off >= fa->fa_size) {
+        return -1;
+    }
+
+    sector->fs_off = off & ~(FLASH_AREA_IMAGE_SECTOR_SIZE - 1);
```
davidvincze, line 391: As I understand it gives an offset of the sector from the beginning of its flash device's storage, however the `struct flash_sector.fs_off` should be an offset value "from the start of its flash area (not device)."

davidvincze: I'm wrong, I realised as soon as I wrote down the above because `off` is relative to the flash area..
Please ignore my comment.

davidvincze (approved):

davidvincze:
FYI flash_area_get_sector() is implemented in upstream: https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/main/bl2/ext/mcuboot/flash_map_extended.c#85

de-nordic:
> FYI flash_area_get_sector() is implemented in upstream: https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/main/bl2/ext/mcuboot/flash_map_extended.c#85

Thanks. I am going through your TFM docs on process and will be doing that soon.
I have been informed that you are in certification or some release process and this will not get in soon, so I have decided to remove MCUboot changes till we get the TFM updated first.

davidvincze:
> > FYI flash_area_get_sector() is implemented in upstream: https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/main/bl2/ext/mcuboot/flash_map_extended.c#85
> 
> Thanks. I am going through your TFM docs on process and will be doing that soon. I have been informed that you are in certification or some release process and this will not get in soon, so I have decided to remove MCUboot changes till we get the TFM updated first.

I mean the `flash_area_get_sector()` function is already available in TF-M.

What would be the target (branch) of your contribution? I think if something needs to be added to the `main` it shouldn't be a problem.

de-nordic:
> FYI flash_area_get_sector() is implemented in upstream: https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/main/bl2/ext/mcuboot/flash_map_extended.c#85

Ah. I missed that. 

I do not need to add it then, it is just not in Zephyr yet. Anyway, since it is not in Zephyr the CI fails on my MCUboot change, so I need for TF-M to be updated on Zephyr first.

tomi-font:
@de-nordic You can simply cherry pick the commit from upstream. We don't need (/want) to update TF-M right now.

de-nordic:
> @de-nordic You can simply cherry pick the commit from upstream. We don't need (/want) to update TF-M right now.

I have no place to cherry pick it. The Zephyr CI fails here, so the cherry pick target would be this repo.

tomi-font:
> I have no place to cherry pick it. The Zephyr CI fails here, so the cherry pick target would be this repo.

Absolutely, the target would be this repo. It's just that you would cherry pick the relevant commit(s) from upstream TF-M as @davidvincze pointed out that what you need is already implemented there.

PR #121, opened 2025-02-17
cmake: set `CMAKE_INSTALL_MESSAGE` in single location

JordanYates:
Only set `CMAKE_INSTALL_MESSAGE` in a single location, remove the unconditional overrides of the value in other locations.

This prevents dozens of rather pointless messages appearing in the build log when `CONFIG_TFM_BUILD_LOG_QUIET` is set.
```
Installing: /home/jordan/code/workspace/build/nrf5340dk/...
````

### CODE REVIEW

**Comment by joerchan** [comment]
@JordanYates Did you submit this patch to the tfm project? It is always better to get them merged there, so that we don’t have to apply this to all future versions of tfm

**Comment by JordanYates** [comment]
> @JordanYates Did you submit this patch to the tfm project? It is always better to get them merged there, so that we don’t have to apply this to all future versions of tfm

https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/35583

**Comment by joerchan** [comment]
Good, I like to keep track of this by adding a reference to the upstream PR to the commit message. Like this:
```
Upstream PR:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/35583
```
This is to help the person that will upgrade to the next version of TF-M find out which patches are no longer needed .

JordanYates:
> Good, I like to keep track of this by adding a reference to the upstream PR to the commit message. Like this:
> 
> ```
> Upstream PR:
> https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/35583
> ```
> 
> This is to help the person that will upgrade to the next version of TF-M find out which patches are no longer needed .

done

joerchan (approved):

tomi-font (changes requested):
Now that the commit is merged upstream, can you cherry pick it directly with `-x` so we also have the commit hash in the commit message? (And if it conflicts just manually add the hash to the message.)

JordanYates:
> Now that the commit is merged upstream, can you cherry pick it directly with `-x` so we also have the commit hash in the commit message? (And if it conflicts just manually add the hash to the message.)

Done

tomi-font (approved):

PR #122, opened 2025-02-20
Initial upstream support for nRF54L15 with TF-M

Vge0rge:
This adds basic support for nRF54L15 with TF-M. 

The support was tested with the tfm_ipc and config_build samples in Zephyr.

All the commits are from upstream TF-M. There is one double revert, this was a  commit only include in this repo and not in upstream TF-M. I reverted here in order to apply cleanly the upstream commits. I don't reapply it since I copied the logic of the commit in the upstream TF-M repo and then cleanly cherry picked the upstream change here.

PR #130, opened 2025-04-14
update to 2.1.2

tomi-font:
- Reverted the local patches/cherry picks.
- Picked all the 2.1.1 => 2.1.2 commits.
- Re-applied all the cherry picks that are only on `main` upstream.
- Lastly, re-applied the local patches.


ozersa:
TF-M v2.1.2 does not have below macro definitions

```
#define SHARED_BOOT_MEASUREMENT_BASE  ...
#define SHARED_BOOT_MEASUREMENT_SIZE ...
#define SHARED_BOOT_MEASUREMENT_LIMIT ...
```

which enabled by below commit on TF-M v2.2.x
https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/152f7866f4e87d105277bd0375d57f8837d6015f%5E%21/

This addition shall be removed from this PR to align with 2.1.2.
Otherwise build error occurs as below
image

tomi-font:
This is one of the commits that were cherry picked as part of https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/122 to bring support for a platform that is not in 2.1 upstream.
Can you indicate how exactly do you encounter this build error? A way to reproduce?

tomi-font:
I dropped the 4 `platform: ext: adi:` commits from this PR, as it's just about updating to 2.1.2 (of which those commits are not part).
I suggest you to open a PR which builds on top of this one (or when this one is merged) to bring support for the platfom you are adding.

wearyzen (commented):
There are a lot of reverts and re-applied patches, which can lead to manual errors..
is there a better way to do this? maybe instead of 2.1.2 we rebase to 2.2.x which has most of these commits and use `git rebase --onto ` to avoid the reverts?

tomi-font:
I would be for using some other strategy than revert, merge, re-apply, but I think that this requires some discussion/approval to actually happen. The current strategy is what has been used before and is allowed: https://docs.zephyrproject.org/latest/develop/modules.html#synchronizing-with-upstream.
Reverting all the local patches beforehand is to make sure that there are no conflicts when bringing in the updated upstream.

About TF-M 2.1/2.2, I think that it's been agreed (in the security WG) to stick to the 2.1 LTS branch for the time being, for certification reasons. Feel free to bring this up in the security WG.

valeriosetti (approved):

wearyzen (approved):

PR #128, opened 2025-04-14
Cherry pick adi/max32657 platform support

ozersa:
This PR fetch commit that pushed after TF-Mv2.1.2 tags, [gerrit/release-2.1.x](https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/refs/heads/release/2.1.x),  [commits here](https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+log/refs/heads/release/2.1.x)

Final commit ("platform: ext: adi: pull in update from TF-M main") comes from [gerrit/main](https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/dd897dee83aeb7be204c4f27ea317743c6f78e65%5E%21/) branch to it build with zephyr/tf-m.

PR #131, opened 2025-04-15
(Zephyr 3.7) update to 2.1.2

tomi-font:
3.7 version of https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/130


ozersa:
@tomi-font thanks for this PR, it covers https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/128 too.
Could you cherry-pick final 4 commits (that I pushed to gerrit/release-2.1.x branch) too on this PR.

SHA:
- a1b21622e35b78c53e2afe2a0e62d5cd5e1142a9
- 3362fd85e60e198c66bd22de9175b9ba87cb612f
- 96174fefb11627fe64f5df92a8d528d22e0db5c4
- 97882fd395325f6b616c430b717fb3ff12c54dbb

image

tomi-font:
>  Could you cherry-pick final 4 commits (that I pushed to gerrit/release-2.1.x branch) too on this PR.

Sure. Though this PR is for 3.7, while your PR targets main. I think you rather want me to do that on https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/130? (BTW a review there would be welcome.)

ozersa:
> > Could you cherry-pick final 4 commits (that I pushed to gerrit/release-2.1.x branch) too on this PR.
> 
> Sure. Though this PR is for 3.7, while your PR targets main. I think you rather want me to do that on #130? (BTW a review there would be welcome.)

I missed to check target branch. You are right. I need it for main branch, https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/130

tomi-font:
> I missed to check target branch. You are right. I need it for main branch, https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/130

added the commits in that PR

nashif:
whats up with all those reverts? why moving to a new version involves so many reverts?

tomi-font:
> whats up with all those reverts? why moving to a new version involves so many reverts?

This is how it's been done, the local patches (which only live in this repo) are reverted, the updated upstream is brought in, and the patches are re-applied. 
Would be happy to adopt a methodology which doesn't require reverts. cc @carlescufi @wearyzen

wearyzen (approved):

carlescufi:
> whats up with all those reverts? why moving to a new version involves so many reverts?

This approach is following what is described in [the guide](https://docs.zephyrproject.org/latest/develop/modules.html#allowed-practices), section "Updating modules by merging the upstream branch". Reverting the commits helps avoiding conflict resolution in the merge commit itself, which is then considered an "evil merge".
The alternative is to revive [this PR](https://github.com/zephyrproject-rtos/zephyr/pull/75330) which proposes something similar to the current approach, but creating a new branch every time we synchronize.

wearyzen:
> > whats up with all those reverts? why moving to a new version involves so many reverts?
> 
> This is how it's been done, the local patches (which only live in this repo) are reverted, the updated upstream is brought in, and the patches are re-applied. Would be happy to adopt a methodology which doesn't require reverts. cc @carlescufi @wearyzen

To add to this, upstream TF-M has LTS version 2.1.x, version 2.2.x and the main branch. Zephyr's TF-M follows the 2.1.x version of upstream (which I heard what was decided in some security group meeting) but also has cherry-picked commits from 2.2.x and main. The current version update is from 2.1.1 to 2.1.2 and reverting/reapplying the patches helps avoid the conflicts.
But it would be good to have better mechanism for such scenarios.

valeriosetti (approved):

nashif:
> This approach is following what is described in [the guide](https://docs.zephyrproject.org/latest/develop/modules.html#allowed-practices), section "Updating modules by merging the upstream branch". Reverting the commits helps avoiding conflict resolution in the merge commit itself, which is then considered an "evil merge".
> The alternative is to revive [this PR](https://github.com/zephyrproject-rtos/zephyr/pull/75330) which proposes something similar to the current approach, but creating a new branch every time we synchronize.

when I see a revert, for me this is always a signal that something bad happened, removing bad commits and such, sure it helps with a clean merge and avoid conflicts but this is not easy to follow when I am looking at an update to a released branch and when trying to understand impact of the changes. 
We need at some point to bite the bullet and rebase the tree and have clean git history with changes on top of upstream clearly visible as commits. This is the only repo IMO where things are being done this way.

PR #140, opened 2025-07-11
platform: stm: disable HW accelerator in BL2

etienne-lms:
Disable HW crypto accelerator in STM BL2 stage since embedding them requires their drivers integrate through the PSA unified driver API. This is a requirement from TF-M since it moved to MCUBoot 2.2.0. We'll be enable to restore them once they comply with PSA driver API.

Note that in mainline TF-M, STM HW accelerator driver are currently not embedded in BL2 firmware so this change provides at least the same embedded feature set as mainline TF-M branch 2.2.x.

We do this by reverting the commit that caused trouble and finally highlight the issue (https://github.com/zephyrproject-rtos/zephyr/issues/92446) and proposing a non invasive commit (regarding TF-M mainline source tree) to ensure HW accelerator drivers are not attempted to be embedded in BL2 firmware.

### DIFF
--- a/platform/ext/accelerator/stm/CMakeLists.txt
+++ b/platform/ext/accelerator/stm/CMakeLists.txt
@@ -58,7 +58,6 @@ if (BL2)
     target_link_libraries(platform_bl2
         PUBLIC
             bl2_crypto_config
-            crypto_service_crypto_hw
     )
 
     target_compile_definitions(bl2_crypto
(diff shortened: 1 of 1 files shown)


nashif (approved):

tomi-font:
@etienne-lms ping

etienne-lms:
Hi, sorry I missed you're post.
We can't submit this P-R to mainline. Mainline does not need `MBEDTLS_ACCELERATOR_PSA_CRYPTO_CONFIG_FILE` (I must crosscheck why) whereas the Zephyr fork did hence this change to add an empty header file instead of just leaving the variable undefined.

etienne-lms:
Note: regarding this change, if you're reworking the commit history, I think commit 65d8615521debd0c4c45153df37bbfe9934e9559 should not be merged. It restores HW crypto accelerator driver in BL2 while ST drivers are not yet ported to PSA driver API in mainline hence cannot be embedded in BL2. Or did I miss something?

tomi-font:
Yeah I was not planning to include 65d8615521debd0c4c45153df37bbfe9934e9559.
You say that mainline "does not need `MBEDTLS_ACCELERATOR_PSA_CRYPTO_CONFIG_FILE`" but could it still be added there?
My point is that if we can get something upstream and drop the downstream-only patch it'd be better this way, so we are not deviating from upstream and maintaining patches. Can some solution be worked out upstream?

PR #143, opened 2025-07-11
BL2: Add mcuboot patch to prevent build warning on FIH_PANIC

etienne-lms:
Embedding mcubootutil `fih_panic_loop()` implementation in TF-M generates a build warning message like the below due to that function missing a `noreturn` attribute:
```
.../arm-none-eabi/include/assert.h: In function '__assert_func':
.../trusted-firmware-m/platform/ext/common/tfm_assert.c:22:1: warning: 'noreturn' function does return
   22 | }
      | ^
```
I've created P-R https://github.com/mcu-tools/mcuboot/pull/2371 to address the issue. Until it is merged, I propose to apply the related patch in Zephyr TF-M repository.

### DIFF
--- a/lib/ext/mcuboot/0006-boot-bootutil-declare-fih_panic_loop-as-noreturn.patch
+++ b/lib/ext/mcuboot/0006-boot-bootutil-declare-fih_panic_loop-as-noreturn.patch
@@ -1,0 +1,59 @@
+From 38eb2499e1b6b4176b76014874cb6253a044d12e Mon Sep 17 00:00:00 2001
+From: Etienne Carriere <etienne.carriere@st.com>
+Date: Thu, 3 Jul 2025 10:36:46 +0200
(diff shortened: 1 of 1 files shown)


nashif:
why are you submitting raw patches instead of fixing the code directly?

etienne-lms:
I'm fine closing this P-R if [#144](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/14)'s commit *"boot_hal: workaround compiler warning"* is the preferred way.

etienne-lms:
> why are you submitting raw patches instead of fixing the code directly?

I took example on patches for other lib/ext/ libs: submitting a P-R to the mainline repo (https://github.com/mcu-tools/mcuboot/pull/2371) and adding a patch file in Zephyr TF-M repo but I understand it's maybe not the preferred way to go.

PR #144, opened 2025-07-11
flash: workaround compiler error with FIH_PANIC

nashif:
Both function are not declared in any headers.

Signed-off-by: Anas Nashif <anas.nashif@intel.com>

### DIFF
--- a/lib/ext/cryptocell-312-runtime/codesafe/src/mbedtls_api/cc_ecp_internal.c
+++ b/lib/ext/cryptocell-312-runtime/codesafe/src/mbedtls_api/cc_ecp_internal.c
@@ -72,6 +72,7 @@
 #define mbedtls_free       free
 #endif
 
+#if defined(ECP_MONTGOMERY)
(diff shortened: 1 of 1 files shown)


tomi-font:
> well, I am not introducing a feature and we need those changes for the 4.2 release. Thats why we have this repo, we should not be waiting for upstream to deal with such issues.

Of course as I said I don't mean to wait for fixes to be merged upstream. But we do cooperate and contribute upstream, that's the preferred way. We aim to have as little local patches as possible because it's a source of issues. So here ideally I would rather send those commits as upstream PRs and cherry pick them here, no need to even have a review on them yet, but at least we have something to point to for future cleanup.

> One of the changes has another PR which has dependencies on mcuboot and is being submitted as a patch already, not sure how this will be dealt with:
>
> https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/143

Hmm not sure either what's the right way to go about this. cc @nordicjm @de-nordic

nashif:
one of the commits submitted upstream https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/40761

PR #147, opened 2025-08-22
platform: ext: common: template: Fix FOTA upgrade

nandojve:
The trusted-firmware-m/31755 changes modified `profile_definition` from 32 to 48 bytes in the platform/ext/common/template/flash_otp_nv_counters_backend.h file.

This change moved fields that are necessary to perform an image upgrade. The issue is visible only when upgrading from an old image (without the changes) to a new one or vice versa. This means that TF-M 2.1.1 do not allow upgrade/downgrade to/from TF-M 2.2.0 or newer versions.

This revert the change to allow migration from Zephyr 3.7.1 LTS to newer images like 4.2.0.

More details from the change can be visualized in: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/31755

### DIFF
--- a/platform/ext/common/template/flash_otp_nv_counters_backend.h
+++ b/platform/ext/common/template/flash_otp_nv_counters_backend.h
@@ -58,7 +58,7 @@ __PACKED_STRUCT flash_otp_nv_counters_region_t {
         uint8_t implementation_id[32];
         uint8_t cert_ref[32];
         uint8_t verification_service_url[32];
-        uint8_t profile_definition[48];
+        uint8_t profile_definition[32];
 
 #ifdef BL2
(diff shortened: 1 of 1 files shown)


nandojve:
Hi @etienne-lms ,

> Indeed this change makes sense to have FW update compatible with previous release of immutable BL2. Since recent TF-M moves to this new field size, maybe it would be worth to have a configuration option to allow selecting the older or newer field size.

I'm waiting the outcome of https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/42130. I recommend someone from ST take a look on it.

nandojve:
Closing this in favor of #152

PR #161, opened 2025-12-03
platform: stm32u5xx: Enable SRAM5 NS region

nandojve:
The stm32u5axx devices have an additional of 832 Kbytes defined as SRAM5, right after SRAM3 region. This region is not enable at moment in the target_cfg and causes a boot loop when system reach the SRAM5 region. Enable SRAM5 region in the Global TrustZone controller (GTZC) to allow NS firmware access SRAM3+SRAM5 regions.

Change-Id: I77ca2111ba27e69c9dfc77b10ae4902d029e6d36
(cherry picked from commit 6b136c622ad2230fd1d67bc6829057acd4e2df71)

### DIFF
--- a/platform/ext/target/stm/common/stm32u5xx/secure/target_cfg.c
+++ b/platform/ext/target/stm/common/stm32u5xx/secure/target_cfg.c
@@ -353,6 +353,7 @@ void sau_and_idau_cfg(void)
 #define MPCBB_LOCK_SRAM2_SIZE 0xf
 #define MPCBB_LOCK_SRAM1_SIZE 0xfff
 #define MPCBB_LOCK_SRAM3_SIZE 0xffffffff
+#define MPCBB_LOCK_SRAM5_SIZE 0xffffffff
 #define MPCBB_LOCK_SRAM4_SIZE 0x1
 #define MPCBB_LOCK(A) MPCBB_LOCK_##A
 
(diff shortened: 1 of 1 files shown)


valeriosetti (approved):

tpambor (commented):
Commit should be prefixed with `[zep fromtree]`. See https://github.com/zephyrproject-rtos/mbedtls?tab=readme-ov-file#additional-patches.

nandojve:
Hi @tpambor ,

> Commit should be prefixed with [zep fromtree]. See https://github.com/zephyrproject-rtos/mbedtls?tab=readme-ov-file#additional-patches.

I see the rational about this but I think this is quite a lot of effort to mark all the commits that came from mainline. Should we just mark the ones that are not from there ? I mean, the mainline ones already have the Change-ID and "cherry picked from commit" which someone needs to check info anyway. Overtime we will have the content as is in mainline with few exceptions.

Well maybe I still don't see the whole logic behind but this will add more complexity in something that is already difficult by nature.

valeriosetti:
Sorry I didn't notice this, but the point raised by @tpambor is correct indeed. @tomi-font recently rewrote the history of this branch adding a tag to all Zephyr custom patches that are applied on top of the official TF-M release.
Full description of available tags and how to use them is [here](https://github.com/zephyrproject-rtos/mbedtls?tab=readme-ov-file#additional-patches).

nandojve:
No problem, I sent the patch yesterday before comment.
If anything still missing let me know and I'll fix ASAP.

valeriosetti (approved):

tomi-font (approved):

tpambor (approved):

PR #169, opened 2026-02-04
[zep noup] Adjust legacy Mbed TLS crypto from PSA Crypto

valeriosetti:
Zephyr is transitioning to PSA Crypto API and removing usage of legacy crypto. Unfortunately in Mbed TLS 3.6 there are still checks (in "check_config.h") that rely on legacy crypto configurations.

Normally in Mbed TLS legacy crypto configurations are auto-adjusted in "config_adjust_legacy_from_psa.h". This wasn't happening in TF-M build because on the NS side MBEDTLS_PSA_CRYPTO_C is not set.

### DIFF
--- a/interface/include/mbedtls/config_psa.h
+++ b/interface/include/mbedtls/config_psa.h
@@ -34,7 +34,7 @@
  * before we deduce what built-ins are required. */
 #include "psa/crypto_adjust_config_key_pair_types.h"
 
-#if defined(MBEDTLS_PSA_CRYPTO_C)
+#if defined(MBEDTLS_PSA_CRYPTO_C) || defined(__ZEPHYR__)
 /* If we are implementing PSA crypto ourselves, then we want to enable the
  * required built-ins. Otherwise, PSA features will be provided by the server. */
 #include "mbedtls/config_adjust_legacy_from_psa.h"


interface/include/mbedtls/config_psa.h
```diff
@@ -34,7 +34,7 @@
  * before we deduce what built-ins are required. */
 #include "psa/crypto_adjust_config_key_pair_types.h"
 
-#if defined(MBEDTLS_PSA_CRYPTO_C)
+#if defined(MBEDTLS_PSA_CRYPTO_C) || defined(__ZEPHYR__)
```
frkv, line 37: This will be problematic in nRF Connect SDK when this is taken downstream

The addition makes legacy APIs available by converting PSA wants algorithms to use the legacy equivalent. This is unfortunately enabling for all crypto features enabled with PSA_WANTS_ regardless of usage and may indeed increase the code size in the NS image.

The effort in this PR is to enable RSA support which is known to lack support in TLS/DTLS and X.509 in Mbed TLS 3.6.x. I suggest the legacy APIs required for RSA support are explicitly set instead. I understand that this is slightly inconvenient since it is using legacy Kconfigs we want to remember, but the fact of the matter is that these configurations are needed for proper TLS/DTLS and X.509 support. 

Please see the similar posting here, as well: https://github.com/Mbed-TLS/mbedtls/issues/9126

valeriosetti:
@frkv thanks for the feedback. I'm closing this now as I addressed the solution differently in https://github.com/zephyrproject-rtos/zephyr/pull/103533.

PR #177, opened 2026-03-26
[zep noup] Do not install Mbed TLS includes in interface dir

valeriosetti:
This is required because upstream Zephyr is now using Mbed TLS 4.x and adding Mbed TLS 3.6.5 includes (which is what it being used in this version of TF-M) would create conflicts due to mixing of versions.

Once TF-M will be bumped to v2.3 this commit can be discarded.

This is integrated and tested upstream by https://github.com/zephyrproject-rtos/zephyr/pull/106258

### DIFF
--- a/cmake/install.cmake
+++ b/cmake/install.cmake
@@ -82,33 +82,35 @@ if (TFM_PARTITION_INTERNAL_TRUSTED_STORAGE)
 endif()
 
 if (TFM_PARTITION_CRYPTO)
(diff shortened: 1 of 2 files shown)


tomi-font (approved):

tomi-font (approved):

frkv (approved):
LGTM

nordicjm (approved):

carlescufi (approved):

PR #191, opened 2026-05-28
[zep noup] Replace tfm_core_panic with psa_panic in __assert_no_args

valeriosetti:
psa_panic is the correct call in this case since it works also in case of high isolation level in TF-M.

### DIFF
--- a/secure_fw/partitions/lib/runtime/assert.c
+++ b/secure_fw/partitions/lib/runtime/assert.c
@@ -11,6 +11,7 @@
 #include "tfm_log_unpriv.h"
 #include "uart_stdout.h"
 #include "tfm_arch.h"
+#include <psa/service.h>
 
 #ifndef __ARMCC_VERSION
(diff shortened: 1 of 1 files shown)


carlescufi (approved):

secure_fw/partitions/lib/runtime/assert.c
etienne-lms: While at it, maybe:
```patch
- * Make it weak so that that acts a fallback implementation.
+ * Make it weak so that it acts as a fallback implementation.

File: secure_fw/partitions/lib/runtime/assert.c [hunk 1]
```diff
@@ -48,5 +48,5 @@ void __assert_puts(const char *msg)
  */
 __WEAK __NO_RETURN void __assert_no_args(void)
 {
-	tfm_core_panic();
+	psa_panic();
```
etienne-lms, line 52: `#include <psa/service.h>` would be welcome I think.

tomi-font (approved):

PR #193, opened 2026-05-29
[zep fromtree] stm32: address wba65xx (and u3c5xx) issues + fix a few u585xx trace messages level

etienne-lms:
Cherry pick commit c630c687aecaba792788b3d0cec17035981e0eb9 to fix https://github.com/zephyrproject-rtos/zephyr/issues/107675.
Cherry pick commit [324f220](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/324f22046176e1eedbce6a7046ab413530cf791f%5E%21/) to fix WBA6 interrupt vector and SAU configuration.
Cherry pick 2 commits fixing levels for a few trace messages on b_u585i_ito02 platform: [7fb282e](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/7fb282eac00be7d8293586d2a09dbda0595aa99e%5E%21/) and [d51edf0](https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/d51edf07af104baa41772af8cf0fa9d51a7430f8%5E%21/).

PR #196, opened 2026-06-01
[zep fromtree] IFX: Updated Infineon platform 

lauracarlesso:
Cherry pick commit from tf-m main https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m.git/+/28b181c7948c2f224bfcb13a6757792be4cc5348 

Additionally, drop previous noup commit that was handled differently in the cherry-picked commit.


lauracarlesso:
@valeriosetti I seem to have messed up somehow and I'd appreciate your help fixing my mess. 

Here's what I did:

- created a new branch from manifest-rev, 
- rebased with -1 HEAD~5 to drop the [noup] commit 132f674ba028105a097ee8ac7474e31a34d856cc 
- cherry-picked (using -x) the one commit I wanted to bring in from the main branch of upstream tf-m which went in cleanly with no conflicts. 

But now the PR is telling me I have conflicts and I don't understand why that is happening. 

I then realized that manifest-rev might not have been at the same commit as the 2.3 branch, so I reabased on top of the 2.3 branch, dropped the [noup] commit and somehow made it worse? 🤣

tomi-font:
You can't just drop a commit that was already merged in `zephyr_tf-m_v2.3.0` because this implies rewriting the history which means conflicts and needing to force-push. What you can do instead is do a revert commit of the noup on top of the latest HEAD, and then add the fromtree. Then whoever updates TF-M next will spot this and drop the original noup as well as its revert.

valeriosetti:
I suspect there is still something wrong with the current git history. The first 3 commits shouldn't be in this PR since they are already part of `zephyr_tf-m_v2.3.0`. The last 2 are OK.
Or am I missing something?

lauracarlesso:
> I suspect there is still something wrong with the current git history. The first 3 commits shouldn't be in this PR since they are already part of `zephyr_tf-m_v2.3.0`. The last 2 are OK. Or am I missing something?

I re-rebased and now it should hopefully be correct with only the two relevant commits in the PR : ) 
Sorry for the confusion!

d3zd3z (dismissed):
You (Laura) must add a Signed-off-by trailer to _both_ of these commits, as you have been involved in the flow.

lauracarlesso:
> You (Laura) must add a Signed-off-by trailer to _both_ of these commits, as you have been involved in the flow.

Added the trailer! 
I thought it was not necessary in the cherry-picks as long as they used the -x option but I'll make sure to add them in future PRs.

valeriosetti:
> You (Laura) must add a Signed-off-by trailer to _both_ of these commits, as you have been involved in the flow.

Mmmm I'm not sure this is strictly required. In this [other PR](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/193) in this repo this is not happening for example. Yes, github complains that some commits are "unverified" but that's only because they are cherry-picked.
IMO that's not a problem because we're cherry-picking changes that landed in upstream TF-M and this repo is also a fork of upstream TF-M, so as long as we check that the `change-Id` matches we should be fine.
@tomi-font wdyt?

valeriosetti (approved):

tomi-font:
Agreeing with @valeriosetti, when writing [the guidelines](https://github.com/zephyrproject-rtos/mbedtls#fromtree) we ended up choosing not to require a sign-off by the person cherry-picking the commit as long as it's a clean cherry-pick with `-x`.

What I think @d3zd3z refers to is some rule that is enforced in Linux contributions, but Zephyr doesn't automatically follow all of Linux's rules.

tomi-font (approved):

PR #199, opened 2026-06-08
[zep noup] toolchain: lower minimum clang version to 19

tpambor:
Lower the minimum required clang version from 20.1.0 to 19.1.0 for the ATfE/clang toolchain. The Zephyr SDK currently ships with clang 19, so enforcing version 20 (which upstream requires since clang 19 reached end-of-life) would break builds using the Zephyr SDK toolchain.

### DIFF
--- a/toolchain_ARMCLANG.cmake
+++ b/toolchain_ARMCLANG.cmake
@@ -8,10 +8,14 @@ cmake_minimum_required(VERSION 3.21)
 
 SET(CMAKE_SYSTEM_NAME Generic)
 
+find_program(CMAKE_C_COMPILER armclang)
+if(CMAKE_C_COMPILER STREQUAL "CMAKE_C_COMPILER-NOTFOUND")
+    message(FATAL_ERROR "Could not find compiler: 'armclang'")
+endif()
+
 # C++ support is not quaranted. This settings is to compile with RPi Pico SDK.
-set(CMAKE_ASM_COMPILER armclang)
-set(CMAKE_C_COMPILER   armclang)
-set(CMAKE_CXX_COMPILER armclang)
+set(CMAKE_ASM_COMPILER ${CMAKE_C_COMPILER})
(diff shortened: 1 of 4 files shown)


tomi-font (changes requested):
please submit these changes upstream instead (and we'd have to ask about Clang 19 vs. 20, maybe 19 is acceptable upstream?)

tpambor:
I cleaned up the patch and submitted it upstream https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/52265.

Also included a revert of https://github.com/zephyrproject-rtos/trusted-firmware-m/commit/228ce6da03930f4a3572e12e3886354f94c5cc42 which never made it to the tree. This is essentially a no-op as `[zep fromlist] Build: use find_program to resolve compiler path` contains the same changes for gcc and extends it also to other compilers and hopefully makes it to tree.

tomi-font (approved):
Looks good, thanks. Let's wait for a bit and see what upstream says.

tomi-font:
Answer I got from upstream about lowering the minimum Clang version to 19:
> ATfE v20 as minimum is because that is the first version publicly/officially available here: https://github.com/arm/arm-toolchain/releases
Yes, there are v18 and v19 but both have linker option incompatibility problem so shall be treated individually.

> FYI: Both v18 and v19 were supported by TF-M originally but cancelled for the reasons I mentioned above. You can add a patch to ATfE toolchain to reenable it by adjusting linker options for removing libc from secure side.

tpambor:
@tomi-font Patch has been merged upstream and I have updated to `[zep fromtree]`

tomi-font (approved):
Please just modify last commit message:

> (which upstream requires ~~since clang 19 reached end-of-life~~)

valeriosetti (approved):

tpambor:
Rebased on top of `zephyr_tf-m_v2.3.0`, no changes.

ithinuel (approved):

tomi-font (approved):

PR #203, opened 2026-06-18
[zep noup] platform: corstone1000: initialise secure partition RW data

npitre:
picolibc's crt0 (the `__PROGRAM_START()` entry used by the corstone1000 `Reset_Handler`) initialises only the single `__data`/`__bss` region. TF-M puts each secure partition's RW data in its own section (LMA != VMA) and relies on the CMSIS scatter `__copy_table__`/`__zero_table__` to initialise them, which newlib's `__cmsis_start()` walked but picolibc does not. The result is that every partition's `.data` stays zero: the ITS `fs_cfg_its.flash_dev` driver pointer is NULL and the Secure Enclave faults the moment Internal Trusted Storage initialises.

This walks both tables in `Reset_Handler` before `__PROGRAM_START()`, restoring the per-partition initialisation. Table symbols are weak so BL1 images that do not emit them still link.

Validated on the Corstone-1000-A320 FVP: with this fix the SE boots TF-M, brings up the host, and Zephyr runs on the Cortex-A320.

### DIFF
--- a/platform/ext/target/arm/corstone1000/Device/Source/startup_corstone1000.c
+++ b/platform/ext/target/arm/corstone1000/Device/Source/startup_corstone1000.c
@@ -22,6 +22,7 @@
  */
 
 #include "tfm_hal_device_header.h"
+#include "tfm_copy_zero_tables.h"
 
 /*----------------------------------------------------------------------------
   External References
@@ -144,5 +145,6 @@ extern const VECTOR_TABLE_Type __VECTOR_TABLE[48];
 void Reset_Handler(void)
 {
     SystemInit();                             /* CMSIS System Initialization */
+    tfm_copy_zero_tables();                   /* picolibc crt0 inits one region only */
     __PROGRAM_START();                        /* Enter PreMain (C library entry point) */
 }
--- a/platform/include/tfm_copy_zero_tables.h
+++ b/platform/include/tfm_copy_zero_tables.h
@@ -1,0 +1,63 @@
+/*
+ * SPDX-FileCopyrightText: Copyright The TrustedFirmware-M Contributors
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef __TFM_COPY_ZERO_TABLES_H__
+#define __TFM_COPY_ZERO_TABLES_H__
+
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct {
+    uint32_t const *src;
+    uint32_t       *dest;
+    uint32_t        wlen;
+} tfm_copy_table_t;
+
+typedef struct {
+    uint32_t *dest;
+    uint32_t  wlen;
+} tfm_zero_table_t;
+
+/* Emitted by the scatter linker scripts; weak so images without them link. */
+extern const tfm_copy_table_t __copy_table_start__[] __attribute__((weak));
+extern const tfm_copy_table_t __copy_table_end__[]   __attribute__((weak));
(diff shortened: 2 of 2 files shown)


platform/ext/target/arm/corstone1000/Device/Source/startup_corstone1000.c
```diff
@@ -141,8 +141,50 @@ extern const VECTOR_TABLE_Type __VECTOR_TABLE[48];
 /*----------------------------------------------------------------------------
   Reset Handler called on controller reset
  *----------------------------------------------------------------------------*/
+/*
+ * picolibc's crt0 (__PROGRAM_START) initialises only the single __data/__bss
+ * region.  TF-M places each secure partition's RW data in its own section
+ * (LMA != VMA) and relies on the CMSIS scatter __copy_table__/__zero_table__
+ * to initialise them - exactly what CMSIS __cmsis_start() did before the
+ * switch to picolibc.  Walk those tables here so per-partition data (e.g. the
```
valeriosetti: Question on this sentence: 

> TF-M places each secure partition's RW data in its own section
> (LMA != VMA) and relies on the CMSIS scatter __copy_table__/__zero_table__
> to initialise them - exactly what CMSIS __cmsis_start() did before the
> switch to picolibc.

As far as you know does it only happens for corstone1000 or it's related to some generic TF-M configuration (ex: isolation level)?
In case of the latter, then I suggest to place this initialization in some more general place.

valeriosetti (changes requested):
The change looks good to me. I only left a question.

platform/ext/target/arm/corstone1000/Device/Source/startup_corstone1000.c
npitre: You're right, it's generic: under picolibc the C-library crt0 initialises only the single __data/__bss region and never walks the scatter __copy_table__/__zero_table__, so any CONFIG_TFM_INCLUDE_STDLIBC platform at isolation >= 2 is affected (rp2350 open-codes the same walk in its extra_init.c). Reworked to a shared tfm_copy_zero_tables() header that the Reset_Handler just calls; rationale is in the commit log.

valeriosetti (approved):
Since this looks like a big problem I was wondering how we ended up not noticing it when bumping to TF-M 2.3. It turns out that the initialization is actually done when `CONFIG_TFM_INCLUDE_STDLIBC` is not set (which is the common case for most of the board in Zephyr I would say):
https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/f187cca1b153c4205efdcd0d231c886a64ed398a/secure_fw/partitions/lib/runtime/crt_start.c#L39

As you mentioned in the commit message RP2350 is doing the same in `extra_init.c` and that commit comes from upstream TF-M, so I don't this issue is strictly related to picolibc since upstream TF-M doesn't use picolibc (that's a Zephyr customization).
A nice follow-up might be to share this piece of code for both corstone1000, RP2350 and all other platforms that are setting `CONFIG_TFM_INCLUDE_STDLIBC`, but that's a TF-M upstream PR and I think it's a bit out of scope here.
It's up to you if you want to try fixing that upstream. For the time being this PR looks good to me. Thanks.

npitre:
@tomi-font this is approved and mergeable whenever you have a moment. It pairs with #204 as the corstone1000 picolibc CI-fix set. Thanks!

npitre:
Superseded by #215, which combines both into a single PR and replaces the picolibc commits with backports of the changes now submitted upstream. Closing.

PR #204, opened 2026-06-18
platform: corstone1000: picolibc build support for the Zephyr SDK toolchain

npitre:
Corstone-1000 is built in Zephyr with the GNU arm-zephyr-eabi toolchain and picolibc, while upstream validates it with Arm Clang. Two `[zep noup]` commits cover the gaps that the GNU+picolibc build hits on this platform:

1. picolibc crt0/stdio support: the BL1 and isolation-S linker scripts gain the picolibc crt0 symbols (matching what the common linker scripts already define), `uart_stdout.c` routes picolibc's `fputch()` hook, and `syscalls_stub.c` provides a weak `_exit()`.
2. a `wrapper.py` fix so `macro_parser` is imported from the script's own directory (it sits next to `wrapper.py`), rather than the parent.

Validated together with the partition-data-init fix (#203) on the Corstone-1000-A320 FVP: the Secure Enclave boots TF-M, brings up the host, and Zephyr runs on the Cortex-A320.

### DIFF
--- a/bl2/ext/mcuboot/scripts/wrapper.py
+++ b/bl2/ext/mcuboot/scripts/wrapper.py
@@ -21,7 +21,7 @@ sys.path = [cwd] + sys.path
 import imgtool
 import imgtool.main
 
-parser_path = os.path.abspath(os.path.join(os.path.dirname(__file__), '../'))
+parser_path = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(parser_path)
 import macro_parser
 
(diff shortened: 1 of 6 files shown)


valeriosetti (approved):

npitre:
@tomi-font approved and mergeable when convenient; pairs with #203 for the corstone1000 picolibc CI fix. Thanks!

tomi-font (commented):
(applies to #203 too) Can't these changes be submitted upstream? For example with some toolchain ifdefs if needed.

Also, if the two PRs are paired together, why aren't they a single PR that would be linked with a single Zephyr PR (which is something we want to test in CI)?

npitre:
Superseded by #215, which combines both into a single PR and replaces the picolibc commits with backports of the changes now submitted upstream. Closing.

PR #209, opened 2026-07-13
Cherry-pick fixes for image encryption support for BL2

tpambor:
This PR cherry-picks upstream patches to fix image encryption support for BL2 (`MCUBOOT_ENC_IMAGES`).

TF-M v2.3.0 always defines `MCUBOOT_USE_PSA_CRYPTO` while it was missed to remove the conflicting option `MCUBOOT_USE_MBED_TLS` before the release, which causes mcuboot to fail to compile: https://github.com/mcu-tools/mcuboot/blob/79d374c1a893aec7af4217428626a7745bb28d49/boot/bootutil/include/bootutil/crypto/aes_ctr.h#L16-L21

In addition, the path for `nist_kw.c` was wrong and resulted in compile failures when compiling with `MCUBOOT_ENCRYPT_KW`.


tomi-font (approved):
Looks good; though we call this cherry-picking and not backporting so I fixed the PR title. :wink:

d3zd3z (approved):

PR #212, opened 2026-07-23
[zep fromtree] platform: nordic_nrf: apply nRF7120 RAM retention directly

erdemsimsek:
This PR brings the upstream changes to the zephyr/trusted-firmware-m
Changes can be seen at https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/52584

Scope of the changes basically

Previously ram-retention for system-off was clear at the system-off phase and then reenabled for the regions defined in the device tree file. Instead of that, ram-retention cleared at the beginning of the application. This allow developers to have configure ram-retention with run-time APIs.
Previously system-off ram retention requests were cached and applied later. Instead of that, applied directly.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
+++ b/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
@@ -172,12 +172,17 @@ if(SOC_NRF7120_TFM_MRAMC_SERVICE)
     )
 endif()
 
-# RAM-control service (set by zephyr's -DNRF_RAM_CTRL_SERVICE=ON).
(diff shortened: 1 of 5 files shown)


tomi-font (changes requested):
`[zep fromtree]` and `git cherry-pick -x` https://github.com/zephyrproject-rtos/mbedtls#additional-patches

erdemsimsek:
> `[zep fromtree]` and `git cherry-pick -x` https://github.com/zephyrproject-rtos/mbedtls#additional-patches

addressed, thank you for your review

tomi-font (approved):

PR #215, opened 2026-07-29
platform: corstone1000: GNU + picolibc build support

npitre:
Supersedes #203 and #204, combined into a single PR as requested.

The GNU + picolibc support went upstream, so the first two commits are backports of the changes now in review there:

- https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/52246 (common)
- https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/52247 (corstone1000 BL1)

A third upstream change (52245, don't force `nano.specs`) does not apply here: this branch has no `nano.specs` link option, that came in after v2.3.0.

The last three commits cannot be upstreamed:

- The full picolibc crt0 symbol set for the secure image, and the CMSIS scatter table walk in the corstone1000 startup. Both exist only because Zephyr's cmsis_6 defines `__PROGRAM_START` as picolibc's `_start` for a picolibc build, bypassing `__cmsis_start()`, which upstream relies on both to initialise `.data`/`.bss` and to walk the copy/zero tables. Without the walk, secure partition RW data placed outside the main region stays uninitialised at isolation level 2 and the enclave faults when Internal Trusted Storage starts. Drop both if cmsis_6 ever stops short-circuiting picolibc.
- The `wrapper.py` `macro_parser` import path fix. Upstream repackaged the bl2 scripts into an installed package, so that bug does not exist there.

This also converges the fork-local picolibc bits in `tfm_common_bl2.ld` and `tfm_isolation_s.ld.template` on the upstream form, so future rebases stay clean.

Verified by building and booting the Zephyr `fvp_corstone1000/a320` board against this branch with no board-local patches: BL1_1 to BL1_2 to BL2 image validated to TF-A BL31 to Zephyr, no secure enclave faults. The matching Zephyr PR bumps the manifest here and removes the band-aid patches.


npitre:
Zephyr side: zephyrproject-rtos/zephyr#114821, which pins the manifest at this PR so both get tested together.

valeriosetti (approved):

tomi-font (changes requested):
`fromtree` is for upstream commits that have been merged, for ongoing reviews it's [`fromlist`](https://github.com/zephyrproject-rtos/mbedtls#fromlist)

tomi-font (commented):
Is there no way all the noup commits would be submitted upstream instead? I don't feel qualified to review them and I'd preferably not have to drag them along forever in this fork.

tomi-font:
Please make sure to give context and explain why every noup needs to be a noup in every commit message.

tomi-font:
@keith-packard Can you review the following commits?
- [[zep noup] platform: common: full picolibc crt0 symbols for the secure image](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215/commits/072b51cf150e105d89e2403872aa9ed7dab10994)
- [[zep noup] platform: add tfm_copy_zero_tables() and use it on corstone1000](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215/commits/13058336e358c603e2a6c8cbe1b10548dc20a014)

platform/ext/common/gcc/tfm_isolation_s.ld.template
```diff
@@ -682,8 +686,13 @@ SECTIONS
         *(SORT_BY_ALIGNMENT(.cy_sharedmem*)) /* Cypress shared memory */
 
         . = ALIGN(TFM_LINKER_TFM_BSS_ALIGNMENT);
+        __tls_base = . ;                /* to satisfy picolibc crt0 */
+        __arm32_tls_tcb_offset = . ;    /* to satisfy picolibc crt0 */
```
keith-packard, line 690: `__arm32_tls_tcb_offset` should be set to the gap the linker will compute between the TLS base register/global value and the start of the TLS area itself -- the compiler/linker make space for a "thread control block" (used by glibc or something?) In any case, for arm32 the minimum TCB size is 8 bytes which will be padded out to the maximum of the alignment of the .tdata and .tbss sections. As it's unlikely that any tls variables will require larger than 8 byte alignment, it would work to simply set this symbol to 8.

keith-packard (changes requested):
`__arm32_tls_tcb_offset` needs to be fixed.

platform/ext/common/gcc/tfm_isolation_s.ld.template
npitre: @keith-packard thanks for catching this. I will set it to 8, but there are a few things I would rather get right at the same time, since this pattern is wider than my patch and you are the authority on it.

The same `= .` assignment is already active in six places in this tree, all in the ATfE scripts that I mirrored when adding the GNU ones: `platform/ext/common/atfe/tfm_common_bl2.ld:137`, `atfe/mps4_corstone3xx_bl1_1.ld:77`, `atfe/mps4_corstone3xx_bl1_2.ld:79`, `atfe/musca_bl2.ld:164`, `atfe/rse_bl1_1.ld:74` and `atfe/rse_bl1_2.ld:75`. Two of them carry the comment "dummy. for picolib crt0 only", so I think the intent was only ever to make the link succeed. Is 8 equally safe for those, or could any of those platforms have `.tdata`/`.tbss` alignment above 8 that would need a larger gap? If it is safe I would rather send a separate change fixing all six than leave the GNU scripts deliberately diverging from the ATfE ones they were copied from. One platform already uses a constant, incidentally: `infineon/common/device/src/atfe/ifx_common_ns.ldc` has `PROVIDE(__arm32_tls_tcb_offset = 0)`.

Second, while I am in there: is `__tls_base = .` placed at the end of `.bss` correct, or does it have a similar problem? I would rather fix both at once than come back to this.

And a broader question, if you do not mind: is there a canonical way to obtain the crt0 contract rather than each platform hand-writing it? Between these scripts there are now several near-copies of the same block defining `__data_start`/`__data_source`/`__data_size`, `__bss_start`/`__bss_size`, `__tls_base`, `__arm32_tls_tcb_offset` and the heap bounds. If picolibc ships a linker fragment that expresses that contract, including it would be better than adding another copy, and I would rather propose that than repeat the pattern.

keith-packard: If you want to avoid any potential issues that might arise from a .tdata or .tbss section with larger alignment restrictions, you can crib from the picolibc sample linker script which uses this pattern:
```
PROVIDE( __tls_align = MAX(ALIGNOF(.tdata), ALIGNOF(.tbss)) );
PROVIDE( __arm32_tls_tcb_offset = MAX(8, __tls_align) );
```
If you also need to deal with arm64 targets, the code is similar, but the size reserved for the TCB on that ABI is 16 bytes instead of 8.

The incorrect value would have broken access to any TLS variables within that project; presumably there aren't any. The only requirement is that a value be provided so that the startup code can initialize the TLS base address with a value; incorrect in this case, but harmless in the absence of any TLS data.

tomi-font:
Now that https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/222 is merged this will need a rebase on top of `zephyr_tf-m_v2.3.1`. I have already changed the PR's base branch to it, please rebase your branch locally and force-push it.

npitre:
@tomi-font @keith-packard rebased on `zephyr_tf-m_v2.3.1` and force-pushed.

There are no `[zep noup]` commits left. The branch is now two `[zep fromtree]`
and two `[zep fromlist]`.

The two `[zep fromlist]` commits are submitted upstream and carry their
Change-Ids:

- `Ib67d81416c4667ca24fa4eaf7b691787fd005de4`, the secure image crt0 symbols
- `I2e4d25e944e4df6095407abfd0e792312222b122`, the shared `tfm_copy_zero_tables()`

On why the secure image needs those symbols at all: the earlier upstream change
claimed it did not, because the SPE reaches the C library through CMSIS
`__cmsis_start()`. That is wrong for picolibc. CMSIS_6 defines
`__PROGRAM_START` as `_start` when `__PICOLIBC__` is set
(ARM-software/CMSIS_6#246), above its own `__cmsis_start` definition, so a
GNU + picolibc build enters crt0 directly and `__cmsis_start()` is never
compiled. The upstream commit message corrects that claim.

@keith-packard on `__arm32_tls_tcb_offset`: I tried the pattern you suggested
and it does not work here. `ALIGNOF()` on a section that does not exist is a
hard link error:

```
ld: undefined section `.tdata' referenced in expression
```

No TF-M linker script defines `.tdata` or `.tbss`, and TF-M uses no
`_Thread_local` or `__thread` anywhere, so there is no alignment to take the
maximum of and a bare constant is the whole answer. Worth adding that the
symbol has to be defined outside the output section: assigning a constant
inside `.bss` still yields a section relative value, so the address-as-size bug
survives.

```
__arm32_tls_tcb_offset = 8;            inside .bss  ->  20000024 B
PROVIDE(__arm32_tls_tcb_offset = 8);   outside      ->  00000008 A
```

I have left the existing `= .` form here to match the nine other sites rather
than fix one of them in passing. As you say it is inert without TLS data. The
cleanup is a separate change covering all of them.

keith-packard:
> I have left the existing `= .` form here to match the nine other sites rather than fix one of them in passing. As you say it is inert without TLS data. The cleanup is a separate change covering all of them.

This seems like a trap though; if anything starts using TLS data, which includes any C library function that exposes errno, then this will have the wrong value. Is there some reason we cannot place the `= 8` form at global scope?

npitre:
@keith-packard OK. I preserved the erroneous symbol addition to mimic what is
already there, then added a commit on top fixing them all at once. It is in
this PR as `[zep fromlist]`, Change-Id I21f26965f160d0163d566b9cfb8baf9a3d74abf9.

npitre:
This branch is what zephyrproject-rtos/zephyr#114821 builds against, and CI
there is green: twister builds the Corstone-1000 board with it, and compliance
passes. The only red is the DNM label that guards manifest changes.

platform/ext/common/gcc/tfm_common_bl2.ld
keith-packard: duplicate symbol?

keith-packard (approved):
Other than one duplicate __arm32_tls_tcb_offset, this looks good.

platform/ext/common/gcc/tfm_common_bl2.ld
npitre: Duplication removed with last push.

keith-packard (approved):

tomi-font (changes requested):
- [[zep fromtree] Platform: common: support building with picolibc](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215/commits/1e9edaee0e58562342f9e427d856b35b072dbf30) conflicts when cherry-picking the upstream commit, so it should be marked `fromtree dirty`

- [[zep fromtree] Platform: Corstone1000: support GNU + picolibc build](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215/commits/bb56318b43fee9a3a1e22ab0dd994c2f1ab8c7be) seems to have extra changes that are not needed and the upstream commit cherry-picks cleanly. If those changes are needed for some reason, they should be in a separate commit.

- The rest looks good.

npitre:
@tomi-font both fixed.

First one is now `[zep fromtree dirty]`. It does conflict on
`tfm_common_bl2.ld` when cherry-picked onto `zephyr_tf-m_v2.3.1`.

Second one is now a plain `cherry-pick -x` of the upstream commit, no extra
changes. The two `PROVIDE(__heap_start/__heap_end)` lines I had removed did
not need their own commit: upstream defines the same symbols after the
section, `PROVIDE` yields to that later assignment, and the values come out
the same either way.

valeriosetti (approved):

npitre:
@tomi-font could you take another look when you have a moment? Both points from
your 4 September review are addressed: the first commit is now tagged
`[zep fromtree dirty]`, and the second is a plain `cherry-pick -x` of the
upstream commit with no extra changes. Keith and Valerio have approved since.

The v4.5 feature freeze starts on 28 September. This is the last gate on the
Zephyr side: zephyrproject-rtos/zephyr#114821 pins the manifest at
`pull/215/head` and cannot merge until this one does, so it needs to land here
with enough margin for the manifest update and that PR to follow. The board has
been in tree since February and disabled in CI since June.

tomi-font:
Sure, sorry for the delay. Please re-request review when a reviewer's comments have been addressed: 
kuva

tomi-font (commented):
- It would make more sense to have `[zep noup] partial revert of "add missing common linker references"` first, which also reverts what makes `Platform: common: support building with picolibc` conflict, that way it could be a clean cherry-pick.
- Rest looks good.
- Please don't forget to follow up with the upstream PRs that are cherry-picked here.

npitre:
Reordered as suggested and rebased on `ef007cd771`.

Both picolibc commits are now plain `[zep fromtree]` cherry-picks, byte identical to upstream `a8feb67260` and `8e8331dd54`, so nothing is tagged `dirty` any more.

Getting there took slightly more than moving commits around. `platform/ext/common/gcc/tfm_common_bl2.ld` had picked up 8 lines from `a08b6c24de7f` that define the same crt0 symbols upstream defines, so the upstream commit could not apply cleanly and two of the local definitions were wrong once upstream's were in place: `__tls_base` pointed into `.heap`, and the `__arm32_tls_tcb_offset` constant was overridden by a later plain assignment. The series now opens with a `[zep noup]` partial revert of that file's hunk, and the rest applies as clean cherry-picks on top. The three other files `a08b6c24de7f` touches are untouched.

On following up upstream, the three `[zep fromlist]` commits are on Gerrit:

- [53947](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/53947) Platform: common: emit crt0 symbols for the secure image
- [53948](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/53948) Platform: add a shared tfm_copy_zero_tables()
- [53950](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/53950) Platform: define __arm32_tls_tcb_offset as a constant

All three are Verified +1 and waiting on review. NXP now has a dependent change ([54186](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/54186)) built on 53948 for four of their platforms, also targeting Zephyr 4.5.

tomi-font:
For example I just opened https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/53947 and saw there is a 2 weeks old review comment from Anton that suggests a change.

tomi-font (approved):
Perfect!

PR #217, opened 2026-08-03
[zep fromtree] platform: nordic_nrf: Allow reading FICR in nRF91

Vge0rge:
In the nRF91 series the FICR is configured as secure and thus it cannot be read directly from the NS app. Add the FICR info addresses to the memory read service so that they can be read by the NS.

Change-Id: I8e85bb830a35922971275cbd16f9055d4552e2a5


tomi-font (changes requested):
`-x`

Vge0rge:
> `-x`

Added, thanks

tomi-font (approved):

PR #223, opened 2026-08-14
[zep fromtree] platform: stm/adi: fix builds without Crypto partition

tomi-font:
The affected platforms enable CRYPTO_HW_ACCELERATOR, which results in the TF_PSA_CRYPTO_CONFIG_FILE
(lib/ext/tf-psa-crypto/tfpsacrypto_config/crypto_config_default.h) to include tf_psa_crypto_accelerator_config.h.

However those platforms added the required include directory only when the Crypto partition was enabled.
Attempting to build would fail in platform/ext/common/provisioning.c which includes psa/crypto.h just to get the value of PSA_ECC_FAMILY_SECP_R1.

Add the include directory to the platform_s target also when the Crypto partition is disabled to fix this, and at the same time fix two of those paths which were erroneous.

### DIFF
--- a/platform/ext/target/adi/max32657/accelerator/CMakeLists.txt
+++ b/platform/ext/target/adi/max32657/accelerator/CMakeLists.txt
@@ -9,6 +9,11 @@
 
 ############################ Crypto Service ####################################
 
+target_include_directories(platform_s
+    PUBLIC
+        ${PLATFORM_DIR}/ext/target/${TFM_PLATFORM}/accelerator/include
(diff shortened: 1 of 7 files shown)


tomi-font:
FYI @etienne-lms

valeriosetti (approved):

Vge0rge (approved):

etienne-lms:
Thanks for the fix. Sorry I miss to provide feedback.

PR #222, opened 2026-08-13
update to v2.3.1

tomi-font:

PR #229, opened 2026-08-26
Bump to 2.3.1

tpambor:
I would suggest to redo #222 on top of the upstream tagged commit for the v2.3.1 release to retain it properly in the history.

I pushed a new base branch `tf-m_v2.3.1` with all commits from upstream 2.3.1 release. Otherwise this is a reopen of #222 against this new base.


tomi-font (approved):
Thanks for the fix!

PR #226, opened 2026-08-21
Support for RSA encrypted images in BL2

tpambor:
Cherry-pick fixes required for support of RSA encrypted images in BL2.

PR #230, opened 2026-09-04
[RFC][zep fromlist] Platform: AN521: Add pseudo-RNG for BL2

tpambor:
The MPS2/AN521 has no hardware TRNG, but BL2 needs random bytes for MbedTLS RSA blinding during private key operations (e.g. RSA-OAEP decryption of encrypted firmware images).

Add a xoshiro128 based pseudo-RNG that provides a strong definition of mbedtls_psa_external_get_random(), overriding the __weak stub in bl2/src/psa_stub_rng.c.

This is NOT suitable for production use. The AN521 is a development target. This PRNG implementation only serves to unblock RSA operations in BL2.


valeriosetti (approved):

tpambor:
This approach was accepted upstream, so I added this back to https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/226 as I deem it worth to be able to test this in CI.

PR #231, opened 2026-09-08
[zep fromlist] Platform: NXP: initialize secure partition data regions

waqar-tahir:
When CONFIG_TFM_INCLUDE_STDLIBC is enabled with the GNU toolchain, the picolibc startup initializes only the primary data and BSS regions. It does not process the CMSIS copy and zero tables describing additional RW and zero-initialized regions.

At isolation level 2 and above, secure partitions can have separate RW data sections. Without processing the CMSIS tables, these regions remain uninitialized and can cause failures when the affected partitions start.

Call tfm_copy_zero_tables() from the secure startup path for the following NXP platforms:

* FRDM-MCXA577
* FRDM-MCXN947
* LPCXpresso55S69
* MCIMX93-EVK

Limit the initialization to secure builds that include the standard C library and use the GNU toolchain.

This change depends on the shared tfm_copy_zero_tables() implementation.

Change-Id: Id942a8d46ce027014bf1c0b94fd3a678852b22c8


waqar-tahir:
This PR depends on zephyrproject-rtos/trusted-firmware-m#215. Once the dependent PR is merged, I will open the corresponding Zephyr PR. Opening it beforehand would result in build failures for the affected NXP platforms.

waqar-tahir:
@valeriosetti @tomi-font Please review this change.
I will open the zephyr manifest update PR as soon as dependent change: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215 is merged.

valeriosetti (approved):

tomi-font (approved):

waqar-tahir:
I have just rebased the PR as dependent PR: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/215 is merged.

